PR #101, opened 2022-03-28
[Feature] Expose new method moveVariables

alexandraFlavia9:
This PR adds a new method called 'moveVariables' used for changing multiple variables positions

## Related tickets

-   [WRS-506](https://support.chili-publish.com/projects/WRS/issues/WRS-506)

### DIFF
--- a/types/VariableTypes.ts
+++ b/types/VariableTypes.ts
@@ -15,3 +15,9 @@ export type Variable  = {
     value?: string;
     defaultValue?: string;
 }
+
+export type MoveVariable = {
+    id: string;
+    parent: string;
+    order: number;
+}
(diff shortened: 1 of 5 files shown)


rizikolik (approved):

types/VariableTypes.ts
```diff
@@ -14,4 +14,10 @@ export type Variable  = {
     occurrences?: number;
     value?: string;
     defaultValue?: string;
+}
+
+export type MoveVariable = {
+    id: string;
+    parent: string;
+    order: number;
 }
```
brapoprod, line 23: shouldnt this EOL be handled by prettier and the pre-commit hook?

brapoprod (approved):

types/VariableTypes.ts
alexandraFlavia9: Nope, we have endOfLine set to auto, which means maintaining existing line endings

PR #140, opened 2022-09-19
[Feature] Update SDK to DEV and PRD pipelines and fixed editor

brapoprod:
This PR updates pipelines to go to correct PRD and DEV environments as well as fixing the editor version in place

## Related tickets

-   [WRS-800](https://support.chili-publish.com/projects/WRS/issues/WRS-800)

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -56,8 +56,8 @@ jobs:
               uses: martinbeentjes/npm-get-version-action@master
             - name: copy file branch
               run: |
-                  path=upload/builds/latest
-                  versionpath=upload/builds/${{ steps.package-version.outputs.current-version}}
+                  path=upload/latest
+                  versionpath=upload/${{ steps.package-version.outputs.current-version}}
                   mkdir -p ${path%"/merge"}
                   cp -R _bundles/* ${path%"/merge"}
                   mkdir -p ${versionpath%"/merge"}
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -66,7 +66,7 @@ jobs:
               uses: lauchacarro/Azure-Storage-Action@v1.1
               with:
                   # Connection String of the Azure Storage Container
-                  connection-string: ${{ secrets.AZURE_CDN_SECRET }}
+                  connection-string: ${{ secrets.AZURE_CDN_STUDIO_DEV_CONNECTION_STRING }}
                   # Folder what contains the files for upload to Storage Container
                   folder: upload
                   # Name of Blob Container Storage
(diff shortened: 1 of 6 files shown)


src/index.ts
```diff
@@ -70,6 +70,8 @@ export { ColorType } from '../types/ColorStyleTypes';
 
 let connection: Connection;
 
+const FIXED_EDITOR_LINK = 'https://stgrafxstudioprdpublic.blob.core.windows.net/editor/0.0.6/web';
```
pietervp, line 73: We want to use the CDN link: https://studio-cdn.chiligrafx.com/editor/0.0.6/web

.github/workflows/publish-release-npm.yml
```diff
@@ -54,7 +54,7 @@ jobs:
               uses: lauchacarro/Azure-Storage-Action@v1.1
               with:
                   # Connection String of the Azure Storage Container
-                  connection-string: ${{ secrets.AZURE_CDN_SECRET }}
+                  connection-string: ${{ secrets.AZURE_CDN_STUDIO_PRD_CONNECTION_STRING }}
                   # Folder what contains the files for upload to Storage Container
                   folder: upload
```
pietervp, line 59: Not sure what this means for the eventual CDN path of the SDK. It should be `<cdn-domain>/sdk/<version>/abc.js` by the looks of it, with current configuration `/builds/` will be part of the path?

pietervp: Also, `blob-container-name` should it be `sdk`? It refers to the container on the root level of the Azure Storage account IIRC

.github/workflows/publish-package.yml
```diff
@@ -66,7 +66,7 @@ jobs:
               uses: lauchacarro/Azure-Storage-Action@v1.1
               with:
                   # Connection String of the Azure Storage Container
-                  connection-string: ${{ secrets.AZURE_CDN_SECRET }}
+                  connection-string: ${{ secrets.AZURE_CDN_STUDIO_DEV_CONNECTION_STRING }}
                   # Folder what contains the files for upload to Storage Container
                   folder: upload
```
pietervp, line 71: Probably same issue as the `publish-release-npm.yml`? We want the CDN URL's to be as short as possible

pietervp (changes requested):

.github/workflows/publish-release-npm.yml
brapoprod: blob-container-name should be correct being SDK, storage account = stgrafxstudioprdpublic => container = SDK.
SA is defined by the connectionstring secret

pietervp (approved):

alexandraFlavia9 (approved):

PR #147, opened 2022-09-29
[Feature] Add configuration SDK methods

pietervp:
This PR

## Related tickets

-   [WRS-1](https://support.chili-publish.com/projects/WRS/issues/WRS-1)


src/controllers/ConfigurationController.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * This method returns the list of pages
+     * @returns
+     */
+    setValue = async (key: WellKnownConfigurationKeys | string, value: string) => {
+        const res = await this.#editorAPI;
+        return res.setConfigValue(key, value).then((result) => getEditorResponseData<null>(result));
+    };
+}
```
brapoprod, line 41: general comment, is there a way of getting and setting all config at once?

brapoprod (commented):
Could you install prettier and the prettier pre-commit hook on husky?

src/controllers/ConfigurationController.ts
pietervp: Not yet, for now if we want to set multiple config items, they need to be seperate calls. We could hide this from integrators by adding SDK methods that support this (internally there would still be multiple calls)

brapoprod (approved):

PR #172, opened 2022-10-19
[Fix] [Blocker] Move authentication controller to ConnectorController.ts

pietervp:
This PR moves the authentication controller from MediaController to ConnectorController because it is a general purpose controller applicable to all connectors

## Related tickets

-   [WRS-1](https://support.chili-publish.com/projects/WRS/issues/WRS-1)

### DIFF
--- a/src/controllers/ConnectorController.ts
+++ b/src/controllers/ConnectorController.ts
@@ -19,6 +19,39 @@ export class ConnectorController {
      * @ignore
      */
     #editorAPI: EditorAPI;
+    authentication: ConnectorAuthenticationController;
+
+    /**
+     * @ignore
+     */
+    constructor(editorAPI: EditorAPI) {
+        this.#editorAPI = editorAPI;
+        this.authentication = new ConnectorAuthenticationController(editorAPI);
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


src/controllers/ConnectorController.ts
```diff
@@ -40,3 +42,50 @@ export class ConnectorController {
...
+     * @param headerName name of the header
+     * @param headerValue value of the header
+     */
+    setHttpHeader = async (connectorId: string, headerName: string, headerValue: string) => {
+        const res = await this.#editorAPI;
+        return res
+            .connectorAuthenticationSetHttpHeader(connectorId, headerName, headerValue)
+            .then((result) => getEditorResponseData<null>(result));
+    };
+}
```
brapoprod, line 91: I would actually prefer to have this in an entirely separate file

brapoprod (changes requested):

src/controllers/ConnectorController.ts
brapoprod: UNLESS this doesn't need to be exported

alexandraFlavia9 (approved):

brapoprod (approved):

PR #183, opened 2022-11-03
[Feature] Connector configuration flow

Dvergar:
**Can be reviewed**

This PR changes the connector configuration flow, this is a breaking-change.

Connector configuration should now be done via the `configure` call.
Any internal configuration call such as `setConnectorOptions`, `setChiliToken`... will only be applied with `updateConnectorConfiguration`.

All of it is wrapped into a configurator callback.

`connectorStateChanged` stream is now exposed. You should check if connector is `loaded` before applying connector configuration.
MediaPanel can also rely on it for proper visual feedback on connector states.

Can be tested against https://agreeable-mud-052c24203.azurestaticapps.net/?editor=EDT-638

## Related tickets

-   [EDT-638](https://support.chili-publish.com/browse/EDT-638)


brapoprod (approved):
Please make sure in the future that breaking changes are communicated well and try to avoid them

pietervp (approved):

PR #197, opened 2022-11-11
[Feature] Adds new method for inserting text variables

alexandraFlavia9:
This PR adds new method for inserting text variables

## Related tickets

-   [WRS-921](https://support.chili-publish.com/projects/WRS/issues/WRS-921)


brapoprod (approved):

abdelhalimkhouas (approved):

pietervp (changes requested):
In the future more methods involving text editing / inserting will need to be added. A couple of options are considered:

`SDK.frame.textEditSession("frameId", function(textApi) { textApi.insertVariableAt(...) })`
or
`SDK.text.insertVariableAt(0, var1)`

On the other hand, we want to add assigning variables to Workspace. This means having access to a SDK method. So to be a more future proof it would be good to add this SDK method to a new controller `experimental` so we can avoid integration relying on this.
This would both unblock workspace and keep our options open.

pietervp (approved):

PR #225, opened 2022-12-29
[FIX] Updated the src property on frames to correctly match the object shapes

srares76:
This PR updates the src property on frame type to correctly match the object shapes

## Related tickets

-   [WRS-854](https://support.chili-publish.com/projects/WRS/issues/WRS-854)

### DIFF
--- a/types/FrameTypes.ts
+++ b/types/FrameTypes.ts
@@ -28,12 +28,23 @@ export type FrameType = {
 };
 
 export type Frame = TextFrame | ImageFrame;
+export type ImageVariableSource = {
+    assetId?: string;
+    connectorId?: string;
+    sourceType: string;
+    variableId: string;
+}
+export type ImageUrlSource = {
+    sourceType: string;
+    url: string;
+}
 // used by new getter methods
+export type FrameSource = ImageVariableSource | ImageUrlSource | undefined;
 export type ImageFrame = {
     frameId: Id;
     frameName: string;
     frameType: FrameTypeEnum.image;
-    src: string;
+    src: FrameSource;
     blendMode: BlendMode;
 };
 export type TextFrame = {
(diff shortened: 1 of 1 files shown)


types/FrameTypes.ts
```diff
@@ -28,12 +28,23 @@ export type FrameType = {
...
 
 export type Frame = TextFrame | ImageFrame;
+export type ImageVariableSource = {
+    assetId?: string;
+    connectorId?: string;
+    sourceType: string;
+    variableId: string;
+}
+export type ImageUrlSource = {
+    sourceType: string;
```
pkgacek, line 38: create an enum ImageSourceType with `url` and `urlVariable` and then use them in the related types as `sourceType`.

types/FrameTypes.ts
```diff
@@ -28,12 +28,23 @@ export type FrameType = {
...
+    url: string;
+}
 // used by new getter methods
+export type FrameSource = ImageVariableSource | ImageUrlSource | undefined;
 export type ImageFrame = {
     frameId: Id;
     frameName: string;
     frameType: FrameTypeEnum.image;
-    src: string;
+    src: FrameSource;
```
pkgacek, line 47: ImageFrameSource

types/FrameTypes.ts
```diff
@@ -28,12 +28,23 @@ export type FrameType = {
 };
 
 export type Frame = TextFrame | ImageFrame;
+export type ImageVariableSource = {
```
pkgacek, line 31: ImageUrlVariableSource

srares76: Isn't it confusing to add Url in the name? I mean it doesn't rely on any url after all

pkgacek: Yea, I get it, but the `sourceType` for that type is `urlVariable` so it would make sense for now to use `ImageUrlVariableSoruce`.

srares76: Ok, agreed

pkgacek (approved):

evan-mcgeek (approved):

PR #226, opened 2022-12-29
[Feature] Add `zoomToPage` on `CanvasController`

evan-mcgeek:
This PR adds a new functionality `zoomToPage()` via new CanvasController to be able to fit the page to the viewport OR the rectangle provided.

## Related tickets

-   [EDT-697](https://support.chili-publish.com/browse/EDT-697)

### DIFF
--- a/src/controllers/ZoomController.ts
+++ b/src/controllers/ZoomController.ts
@@ -29,7 +29,7 @@ export class ZoomController {
      * @param height
      * @returns
      */
-    zoomToFit = async (
+    zoomToPage = async (
         pageId?: Id | null,
         left?: number | null,
         top?: number | null,
@@ -37,6 +37,6 @@ export class ZoomController {
         height?: number | null,
     ) => {
         const res = await this.#editorAPI;
-        return res.zoomToFit(pageId, left, top, width, height).then((result) => getEditorResponseData<null>(result));
+        return res.zoomToPage(pageId, left, top, width, height).then((result) => getEditorResponseData<null>(result));
     };
 }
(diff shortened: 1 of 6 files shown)


pkgacek (approved):

Dvergar (changes requested):
LGTM but I requested a change and/or a feedback on engine-side which would have an effect on this PR, just blocking it to prevent a merge.

See: https://github.com/chili-publish/studio-engine/pull/36

src/controllers/ZoomController.ts
```diff
@@ -0,0 +1,42 @@
...
+        pageId?: Id | null,
+        left?: number | null,
+        top?: number | null,
+        width?: number | null,
+        height?: number | null,
+    ) => {
+        const res = await this.#editorAPI;
+        return res.zoomToFit(pageId, left, top, width, height).then((result) => getEditorResponseData<null>(result));
+    };
+}
```
brapoprod, line 42: Isn't it a bit overkill to have a zoomcontroller?
Doesn't it make sense to put all canvas related stuff under a CanvasController?
Maybe there we can make a distinction f.e.
SDK.canvas.zoom.zoomToFit?

Dvergar: I'll leave @evan-mcgeek give its input but we discussed it several times and I think we agreed that there would be a discussion following the SDK PR.
We hesitated between putting it beside a zoom controller, canvas controller (or even a page controller).

I myself don't have a strong opinion, it's probably overkill for now, not sure how many zoom functions we are going to get (a second one is coming for general canvas zoom). Canvas controller looks the most generic.

_As a side-note, it's gonna be called zoomToPage._

@evan-mcgeek what do you think ?

evan-mcgeek: I'd leave it separated IMO, considering some possible additional functionality in the future

evan-mcgeek: so what would you say @brapoprod ?

brapoprod (approved):

Dvergar (approved):

PR #229, opened 2022-12-29
[Feature] Add `setImageSource` to `FrameController`

evan-mcgeek:
This PR adds private method `updateImageSource` to FrameController to be able to set or reset the image source for the ImageFrame using the a single method exposed via Editor API

## Related tickets

-   [EDT-693](https://support.chili-publish.com/browse/EDT-693)

### DIFF
--- a/src/controllers/ExperimentController.ts
+++ b/src/controllers/ExperimentController.ts
@@ -1,5 +1,6 @@
 import { EditorAPI, Id } from '../../types/CommonTypes';
 import { getEditorResponseData } from '../utils/EditorResponseData';
+import { ImageSourceTypeEnum, ImageUrlVariableSource } from '../../types/FrameTypes';
 
 /**
  * The ExperimentController contains all SDK functions that are considered for addition,
@@ -28,7 +29,10 @@ export class ExperimentController {
      */
     insertImageVariableToFrame = async (imageFrameId: Id, variableId: string) => {
         const res = await this.#editorAPI;
-        return res.assignImageVariable(imageFrameId, variableId).then((result) => getEditorResponseData<null>(result));
+        const src: ImageUrlVariableSource = { variableId: variableId, sourceType: ImageSourceTypeEnum.urlVariable };
+        return res
+            .setImageSource(imageFrameId, JSON.stringify(src))
+            .then((result) => getEditorResponseData<null>(result));
     };
 
     /**
--- a/src/controllers/ExperimentController.ts
+++ b/src/controllers/ExperimentController.ts
@@ -1,6 +1,5 @@
 import { EditorAPI, Id } from '../../types/CommonTypes';
 import { getEditorResponseData } from '../utils/EditorResponseData';
-import { ImageSourceTypeEnum, ImageUrlVariableSource } from '../../types/FrameTypes';
 
 /**
  * The ExperimentController contains all SDK functions that are considered for addition,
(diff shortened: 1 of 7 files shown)


pkgacek (approved):

types/DocumentTypes.ts
```diff
@@ -68,6 +68,9 @@ export enum ImageFrameSourceType {
     assetProvider = 'assetProvider',
     variable = 'variable',
 }
+
+export type ImageSource = ImageFrameSource | UrlImageFrameSource | null;
```
Dvergar, line 72: It already exists in `frameTypes.ts` as `ImageFrameSource`.

`export type ImageFrameSource = ImageUrlVariableSource | ImageUrlSource | ImageConnectorVariableSource;`

I noted that It is missing `ImageAssetProviderSource`.

also looking at `ImageConnectorVariableSource`, `variableId` property is missing as well.

In any case, if you are wondering why we have subtypes of variable in there it's probably because we shouldn't. I made a ticket for it https://support.chili-publish.com/browse/EDT-758

I vote to align on that one when you're all back from holidays 🎄.

src/controllers/FrameController.ts
```diff
@@ -410,6 +411,17 @@ export class FrameController {
         return res.assignImageFromUrl(imageFrameId, url).then((result) => getEditorResponseData<null>(result));
     };
 
+    /**
```
Dvergar, line 414: It is very clear but in this case I would call it `removeImageFrameSource` (unless other suggestion) and passing null as a source. `setImageSource` should probably not be exposed as such.

To WRS: the idea (from @Matthiee) is to expose more general functions from the engine and splitting it on SDK-side, this method should cover `assignImage`, `assignImageFromUrl`, `assignImageVariable` and the missing `removeImageFrameSource` (already implemented function can be refactored later).

Dvergar (changes requested):

src/controllers/FrameController.ts
Matthiee: Like it was done [here](https://github.com/chili-publish/editor-sdk/pull/219/files#diff-7639a5d7174e6ff877f3b1ab6e7304b200dd8b0eb33f5319157d837cff11192eR193) seems to be the best solution imo.

types/DocumentTypes.ts
evan-mcgeek: I've noticed there is a `url` property in `ImageConnectorVariableSource`. Not sure if it should be there

src/controllers/ExperimentController.ts
```diff
@@ -28,7 +29,10 @@ export class ExperimentController {
      */
     insertImageVariableToFrame = async (imageFrameId: Id, variableId: string) => {
         const res = await this.#editorAPI;
-        return res.assignImageVariable(imageFrameId, variableId).then((result) => getEditorResponseData<null>(result));
+        const src: ImageUrlVariableSource = { variableId: variableId, sourceType: ImageSourceTypeEnum.urlVariable };
```
Dvergar, line 32: That url (temporary) specific source was a suggestion by me - as for some reason - I thought it was the only use WRS did of `insertImageVariableToFrame` but it actually seems to not be used at all...

I'm not sure if @srares76 or @pkgacek  are using it in a current ticket but maybe we should revert back to `assignImageVariable` for now and refactor it when it's time for it > [EDT-758](https://support.chili-publish.com/browse/EDT-758). Unless you think otherwise Evan, i'd say it's the way to go to avoid any issue.

Dvergar (changes requested):
Have a look at my comment otherwise LGTM.

src/controllers/ExperimentController.ts
evan-mcgeek: what's about this comment now @Dvergar ?

evan-mcgeek: @brapoprod maybe you have some thoughts one that one? 
Basically the next ticket will fix this issue, but what will be the best option for now?

evan-mcgeek: UPD

Resolved for now by reverting back `assignImageVariable(imageFrameId, variableId)` method on the editor side to avoid any breaking changes atm

Matthiee (approved):

Dvergar (approved):

brapoprod (approved):

pkgacek (approved):

PR #235, opened 2023-01-10
[Fix] Clean up connector events and states

Matthiee:
This PR splits the connector states into events and states. 

## Related tickets

-   [EDT-666](https://support.chili-publish.com/projects/EDT/issues/EDT-666)

### DIFF
--- a/src/controllers/SubscriberController.ts
+++ b/src/controllers/SubscriberController.ts
@@ -191,6 +191,7 @@ export class SubscriberController {
      * - ready
      * - reloaded
      * - error
+     * - authentication changed
      * @param connectorEvent Stringified object of ConnectorEvent
      */
     onConnectorStateChanged = (connectorEvent: string) => {
--- a/src/controllers/SubscriberController.ts
+++ b/src/controllers/SubscriberController.ts
@@ -188,12 +188,12 @@ export class SubscriberController {
      * - authChanged
      * - unloaded
      * - stateChanged
-     * 
+     *
      * stateChanged - this event will be triggered by connector states: loading, loaded, running, ready, error
      * @param connectorEvent Stringified object of ConnectorEvent
      */
-    onConnectorStateChanged = (connectorEvent: string) => {
-        const callBack = this.config.onConnectorStateChanged;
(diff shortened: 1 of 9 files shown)


pkgacek (approved):

src/controllers/SubscriberController.ts
```diff
@@ -191,6 +191,7 @@ export class SubscriberController {
      * - ready
```
Dvergar, line 191: _(I still don't know how to review on expanded code, google is not helping)_

I feel the description is confusing now, the listener is returning events that are indeed triggered by those states but only available from a `getState` call and not actually related to all of those events.

Something along those line might be better ?

> Listener on connector events, this listener will get triggered when a connector emits one of those events
> * `reloadRequired`
> * `authChanged`
> * `unloaded`
> * `stateChanged`

Maybe adding `stateChanged - this event will be triggered by connector states: loading, loaded, running, ready, error` ?

And since WRS does not seem to use `onConnectorStateChanged` directly maybe it would be good to rename it `onConnectorChanged` ? seems like it's a switching connectors stream. `onConnectorEvents` ? but the `changed` suffix seems mandatory. Any thought on that ?

It's just that we have `stateChanged` as an event type and the listener is called `onConnectorStateChanged` which is accurate but confusing, we kind of expect to get those loading, loaded... states instead of the only 4 events.

Dvergar (changes requested):

src/controllers/SubscriberController.ts
Matthiee: I first also wanted to go with `onConnectorEvents` but since it would be the only one not ending on `Changed` I decided to not do it. @brapoprod any opinion on the naming here? I feel like `onConnectorEvents` makes more sense.

brapoprod: It should describe what it is, if it makes more sense to diverge from what is there in other calls, that makes sense if the content is different. So fine f me :)

pkgacek (approved):

Dvergar (approved):
🚀

brapoprod (approved):

PR #237, opened 2023-01-20
[Feature] [EDT-758] ImageFrameSource should not be aware of the variable type

evan-mcgeek:
This PR deprecates the usage of `ImageUrlVariableSource` and `ImageConnectorVariableSource` and combines them to `ImageVariableSource`

## Related tickets

-   [EDT-758](https://support.chili-publish.com/browse/EDT-758)


brapoprod (approved):

types/FrameTypes.ts
```diff
@@ -29,31 +29,20 @@ export type FrameType = {
...
     assetId: string;
     connectorId: string;
     sourceType: ImageSourceTypeEnum.connector;
 };
-export type ImageFrameSource =
-    | ImageUrlVariableSource
-    | ImageUrlSource
-    | ImageConnectorVariableSource
-    | ImageConnectorSource;
+export type ImageFrameSource = ImageUrlSource | ImageConnectorSource;
```
Dvergar, line 45:
```suggestion
export type ImageFrameSource = ImageUrlSource | ImageConnectorSource | ImageVariableSource;
```
Not sure if that's to be in scope of this ticket but beside `ImageUrlSource`, removing the two other types does NOT make tests fail.

I'd say even the tiniest test to lock those expected types somewhere would be good to not accidentally remove them.

Dvergar (changes requested):

types/FrameTypes.ts
evan-mcgeek: my point was to remove the Variable type, since it's not being used by the FrameController now

Dvergar (approved):
Yes tests!❤️

_(There is a warning in experimental controller `Using exported name 'SDK' as identifier for default export`)_

Dvergar (approved):

pkgacek (approved):

types/FrameTypes.ts
```diff
@@ -85,8 +74,7 @@ export type TextFrame = {
 
 export enum ImageSourceTypeEnum {
     url = 'url',
-    urlVariable = 'urlVariable',
-    connectorVariable = 'connectorVariable',
+    variable = 'variable',
     connector = 'assetProvider',
```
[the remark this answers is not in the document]
evan-mcgeek: this is changed because of the EditorAPI changes, `connector` is the right name, however, to avoid any breaking changes, we preferred to keep `assetProvider` value as well

PR #251, opened 2023-02-21
[Fix] Add missing isConstrained & Shape model

Dvergar:
This PR adds a missing `isConstrained` property and adds shape frame type.

### DIFF
--- a/types/FrameTypes.ts
+++ b/types/FrameTypes.ts
@@ -23,11 +23,13 @@ export type FrameType = {
     frameId: Id;
     frameName: string;
     frameType: FrameTypeEnum;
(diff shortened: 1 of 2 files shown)


types/FrameTypes.ts
```diff
@@ -23,11 +23,13 @@ export type FrameType = {
     frameId: Id;
     frameName: string;
     frameType: FrameTypeEnum;
+    // `imageUrl` is not generic: should be removed from model
```
Dvergar, line 26: I was a bit shy to remove this one myself as it may break stuff on WRS.

Matthiee (approved):

abdelhalimkhouas (approved):

PR #263, opened 2023-03-19
Add missing attributes to the ShapeFrame type

abdelhalimkhouas:
This PR Adds missing attributes to the ShapeFrame type

## Related tickets

-   [WRS-963](https://support.chili-publish.com/projects/WRS/issues/WRS-963)

### DIFF
--- a/types/FrameTypes.ts
+++ b/types/FrameTypes.ts
@@ -63,7 +63,6 @@ export type ImageFrame = {
 export type ShapeFrame = {
     frameId: Id;
     frameName: string;
-    shapeType: ShapeType;
     frameType: FrameTypeEnum.shape;
     blendMode: BlendMode;
     constrainProportions: boolean;
@@ -74,7 +73,9 @@ export type ShapeFrame = {
         strokeWeight: number;
         strokeColor: ColorUsage;
     };
-    src: { shapeType: ShapeType };
+    src: { 
+        shapeType: ShapeType 
+    };
 };
 
 export type TextFrame = {
--- a/types/FrameTypes.ts
+++ b/types/FrameTypes.ts
@@ -1,4 +1,5 @@
 // FramePropertiesDto
+import { ColorUsage } from './ColorStyleTypes';
 import { Id, PropertyState } from './CommonTypes';
 
 export type FrameLayoutType = {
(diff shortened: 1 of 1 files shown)


types/FrameTypes.ts
```diff
@@ -62,10 +63,19 @@ export type ImageFrame = {
...
     blendMode: BlendMode;
     constrainProportions: boolean;
+    shapeProperties: {
+        enableFill: boolean;
+        fillColor: ColorUsage;
+        enableStroke: boolean;
+        strokeWeight: number;
+        strokeColor: ColorUsage;
+    };
+    src: {
```
Dvergar, line 76:
```suggestion
    src: ShapeFrameSource
```
I was reading the ticket and it might not be in scope of it but at least for future reference:

We kind of missed the `ShapeFrameSource`.

```typescript
export type ShapeFrameRect = {
    shapeType: ShapeType.rectangle;
};

export type ShapeFrameEllipse = {
    shapeType: ShapeType.ellipse;
};

export type ShapeFramePolygon = {
    shapeType: ShapeType.polygon;
    sides: number;
};

export type ShapeFrameCustom = {
    shapeType: ShapeType.custom;
    width: number;
    height: number;
    drawCommands: any[];
};

export type ShapeFrameSource = ShapeFrameRect | ShapeFrameEllipse | ShapeFramePolygon | ShapeFrameCustom;

export enum ShapeType {
    ellipse = 'ellipse',
    rectangle = 'rectangle',
    polygon = 'polygon',
    custom = 'custom',
}
```

That would be the types but we most probably don't want to expose custom shapes yet though. Ideally I would expose everything but custom shapes do you know more about that @evan-mcgeek ?

evan-mcgeek: I agree with that, since different shapes will have different properties and simply specifying the type of the shape  won't be sufficient anymore. 
But it's definitely not in a scope of this ticket, for that we'd need a separated one to refactor both SDK and Engine approaches.

Dvergar: Thanks for the feedback, I'll auto-resolve the comment then :)

pkgacek (approved):

brapoprod (approved):

PR #266, opened 2023-03-31
Add definition files for Connectors and Actions + Move /types to /src/types

pietervp:
* Moved types to src/types
* Added defintion files for connectors
* Create definition file for actions

The rationale behind this is that the existing `/types` folder was somehow misplaced under the root folder. These concrete types are part of the `src` of this repository and their location in the tree should reflect that. If consumers of SDK did not import the types 'directly' (`import ... from '@chili-publish/studio-sdk/types/commontypes'`), this change should be transparent for them. Although any place that abused the Ts imports will be exposed after this PR gets merged.

Also generated documentation will change slightly, see changes in `tsconfig.json`.

The root `types/` folder is meant to expose Ts definition files. These definition files define functionality implemented elsewhere, in this case in the plugin system of the engine. These definitions define all functionality that will be injected in the QuickJS runtime context, either when using actions, or implementing a connector. By exposing these type definitions, we enable code completion when writing said custom actions / connectors.


types/Actions.d.ts
```diff
@@ -0,0 +1,134 @@
...
+        width: number;
+        height: number;
+    }
+
+    enum FitMode {
+        fill,
+        fit,
+    }
+
+    interface AlternateLayout{
```
Matthiee, line 57: This misses some properties, but it will use the inherited layout instead of the one saved in the document in a later version.

types/Actions.d.ts
```diff
@@ -0,0 +1,134 @@
...
+          readonly layoutId: string,
+          readonly layoutName: string,
+          readonly width: number,
+          readonly height: number,
+          readonly animated: boolean,
+    }
+
+    interface VariableSource {
```
Matthiee, line 65: Image variables can have a source defined as well see `variable_source.dart`.

brapoprod (dismissed):

types/Actions.d.ts
pietervp: @Matthiee I skipped a couple of properties on purpose, and tried to stick to the basics. Do you think we can start with the list in the current PR?

Matthiee: @pietervp sounds good

Matthiee (approved):

brapoprod (approved):

PR #280, opened 2023-05-18
[EDT-970] enter / exit crop mode notifications

evan-mcgeek:
This PR adds a subscription to track entering / exiting the crop mode by exposing `frameId?`

## Related tickets

-   [EDT-970](https://support.chili-publish.com/browse/EDT-970)

### DIFF
--- a/src/controllers/SubscriberController.ts
+++ b/src/controllers/SubscriberController.ts
@@ -1,4 +1,4 @@
-import { ConfigType } from '../types/CommonTypes';
+import { ConfigType, Id } from '../types/CommonTypes';
 import { ToolType } from '../utils/enums';
 
 /**
@@ -87,7 +87,6 @@ export class SubscriberController {
         const callBack = this.config.onDocumentLoaded;
         callBack && callBack();
     };
-
 
     /**
      * To be implemented, gets triggered when clicking on the pageTitle on the canvas.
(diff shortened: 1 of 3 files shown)


src/controllers/SubscriberController.ts
```diff
@@ -242,4 +241,13 @@ export class SubscriberController {
         const callBack = this.config.onShapeCornerRadiusChanged;
         callBack && callBack(JSON.parse(cornerRadius));
     };
+
+    /**
+     * Listener of editor entering / exiting the crop mode
+     * @param id frame id when entering / null when exiting
+     */
+    onCropActiveFrameIdChanged = (id?: Id) => {
```
Dvergar, line 249: Not really a request but just a thought:

The API is nice to me internally but i'm wondering integrator-ux-wise if mapping to:

`onCropModeEnter = (id: Id)` and `onCropActiveFrameExit = ()` would fit better.

Dvergar: Had a call w/ @evan-mcgeek about that, this is obviously an open question but:
- 2 different events for this might not make the best sense
- With this change we would have 3 crop-related streams

pkgacek (approved):

pkgacek (approved):

NicolaVerbeeck (approved):

PR #293, opened 2023-06-07
[Fix] standardize sdk methods

pkgacek:
This PR makes some changes to the API naming convention

## Related tickets

-   [WRS-1174](https://support.chili-publish.com/browse/WRS-1174)


srares76 (approved):

src/controllers/ActionController.ts
```diff
@@ -21,80 +21,81 @@ export class ActionController {
...
     /**
-     * This method returns an action by id
-     * @param actionId The ID of a specific action
-     * @returns
+     * This method returns an action by the id
+     * @param actionId the id of a specific action
+     * @returns action details
      */
-    getActionById = async (actionId: Id) => {
+    getById = async (actionId: Id) => {
```
abdelhalimkhouas, line 36: If i recall correctly we agreed to use `id` instead of `[prefix]Id` pattern

abdelhalimkhouas (changes requested):
for `ActionController` and `CharachterStyleController` you are using `[prefix]Id` pattern as a param, and in other controllers you are using `id` this is one of the inconsistencies that should be tackled here, and I believe we chose to go with `id` @Matthiee can confirm the choice

pkgacek:
> for `ActionController` and `CharachterStyleController` you are using `[prefix]Id` pattern as a param, and in other controllers you are using `id` this is one of the inconsistencies that should be tackled here, and I believe we chose to go with `id` @Matthiee can confirm the choice

You're totally right, I must have missed this!

I would only keep [prefix]Id in case we need to provide two ids to the method to make sure end user is not confused in which order it needs to be used

src/types/DocumentTypes.ts
```diff
@@ -28,31 +28,31 @@ export interface ChiliDocument {
...
+    id: string;
+    name: string;
+    type: FrameTypeEnum;
     blendMode: BlendMode;
     constrainProportions: boolean;
 }
 
 export interface ImageFrame extends DocumentFrame {
-    src: VariableImageFrameSource | AssetProviderImageFrameSource | UrlImageFrameSource;
+    src: VariableImageFrameSource | ConnectorImageFrameSource | UrlImageFrameSource;
```
Dvergar, line 51: Not commenting the connector change but the three sources naming is inconsistent with `FrameType.ts`.

`export type ImageFrameSource = ImageFrameUrlSource | ImageFrameConnectorSource | ImageFrameVariableSource;`
If team is ok for changing this I personally prefer the `FrameType.ts` version.

Dvergar:
In https://github.com/chili-publish/studio-sdk/pull/293/files#diff-051988f5f947f1085f7226e6230d8b5a8cb76d072024a7eff3ded37143c759d0L55 I missed that one 

`slideDirection: SlideDirections;` > `direction: SlideDirections;`

Dvergar (changes requested):

src/types/DocumentTypes.ts
```diff
@@ -113,21 +113,21 @@ export interface TopLayout extends Layout {
     frameAnimations: FrameAnimation[];
     timelineLengthMs: number;
     animated: boolean;
-    parentLayoutId: Id;
+    parentId: Id;
```
Dvergar, line 116: To remove.

src/README_OLD.md
```diff
@@ -190,14 +190,14 @@ window.SDK.tool.setHandTool();
...
 ```
 
 <br/>
 
 ### Recap
 
-So to recap, you first need to initialise the SDK and Editor, then add an empty div with the ID that you want to pass through the SDK to inject the Editor in.
+So to recap, you first need to initialise the SDK and Editor, then add an empty div with the id that you want to pass through the SDK to inject the Editor in.
 Then you're ready to go, if you initialise some listeners with callbacks or use a getter you can easily retrieve data on change or on request.
 To send data you just call the sdk method, with or without parameters (depends on the method).
```
brapoprod, line 203: this file can actually be removed

brapoprod (approved):

abdelhalimkhouas (approved):

brapoprod (approved):

PR #295, opened 2023-06-07
[EDT-910] - adapting abstract colors to switch to replace studio color

olegsolodukhin:
This PR

## Related tickets

-   [EDT-910] - adapting abstract colors to switch to replace studio color - alpha should be in range 0-1.0


Dvergar:
Don't forget to change `hls = 'hls';` to `hsl = 'hsl';` in `\src\types\ColorStyleTypes.ts`

Dvergar (changes requested):
Don't forget to change `hls = 'hls';` to `hsl = 'hsl';` in `\src\types\ColorStyleTypes.ts`

Dvergar (approved):

PR #298, opened 2023-06-08
[Feature] Add List Variable

Matthiee:
This PR adds support for `ListVariable`

## Related tickets

-   [EDT-982](https://support.chili-publish.com/browse/EDT-982)

### DIFF
--- a/src/controllers/VariableController.ts
+++ b/src/controllers/VariableController.ts
@@ -101,6 +101,20 @@ export class VariableController {
     };
 
     /**
+     * This method sets the items of the list variable
+     * 
+     * The items need to be unique and are not case sensitive. 
+     *
+     * @param variableId the id of the list variable
+     * @param items the items of the list
+     * @returns
+     */
+    setListVariableItems = async (variableId: string, items: string[]) => {
+        const res = await this.#editorAPI;
+        return res.setListVariableItems(variableId, items).then((result) => getEditorResponseData<null>(result));
+    };
+
+    /**
      * This method sets a new value for a variable
      *
      * @param variableId the id of the variable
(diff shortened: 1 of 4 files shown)


src/controllers/VariableController.ts
```diff
@@ -100,6 +100,20 @@ export class VariableController {
...
+    /**
+     * This method sets the items of the list variable
+     * 
+     * The items need to be unique and are not case sensitive. 
+     *
+     * @param variableId the id of the list variable
+     * @param items the items of the list
+     * @returns
+     */
+    setListVariableItems = async (variableId: string, items: string[]) => {
```
pkgacek, line 112: Maybe we could rename it to "setListVariable" to be in the same style as "setImageVariable" or "set<TYPE>Variable"? Appending "items" seems redundant as ListVariable is a list of elements (in this case items). Just a thought taking into the account the SDK rewrite in my pr.

src/controllers/VariableController.ts
```diff
@@ -100,6 +100,20 @@ export class VariableController {
...
+    /**
+     * This method sets the items of the list variable
+     *
+     * The items need to be unique and are not case sensitive.
+     *
+     * @param variableId the id of the list variable
+     * @param items the items of the list
+     * @returns
+     */
+    setListVariable = async (variableId: string, items: string[]) => {
```
pkgacek, line 112: VariableId -> id

pkgacek (changes requested):

src/controllers/VariableController.ts
```diff
@@ -100,6 +100,20 @@ export class VariableController {
...
+    /**
+     * This method sets the items of the list variable
+     *
+     * The items need to be unique and are not case sensitive.
+     *
+     * @param id the id of the list variable
+     * @param items the items of the list
+     * @returns
+     */
+    setListVariable = async (id: string, items: string[]) => {
```
pkgacek, line 112: id: Id (sorry, forgot about this one!) If items are also ids, then items: Id[].

pkgacek (approved):

src/controllers/VariableController.ts
Matthiee: **Oops** indeed, however the items are not `Id`'s.

brapoprod (approved):

PR #313, opened 2023-06-28
[Feature] Add boolean variable type

Dvergar:
This PR adds the new boolean variable type.

Can be tested against https://studio.chiligrafx-dev.com/bench.html?workspace=latest&engine=293

## Related tickets
EDT-983
WRS-1201
WRS-1202

### DIFF
--- a/src/types/VariableTypes.ts
+++ b/src/types/VariableTypes.ts
@@ -33,21 +33,23 @@ export interface Variable {
     isVisible: boolean;
     isReadonly: boolean;
     isRequired: boolean;
+    occurrences: number;
 }
 
 export interface ImageVariable extends Variable {
-    occurrences: number;
     src?: ImageVariableSource;
 }
 
 export interface ListVariable extends Variable {
-    occurrences: number;
     items: string[];
     selected?: string;
 }
 
+export interface BooleanVariable extends Variable {
+    value: boolean;
+}
+
 export interface ShortTextVariable extends Variable {
-    occurrences: number;
     value: string;
 }
 
--- a/types/Actions.d.ts
+++ b/types/Actions.d.ts
@@ -111,7 +111,7 @@ declare module 'grafx-studio-actions' {
     }
 
     /** Respresents a Variable inside Actions */
-    type Variable = ShortTextVariable | LongTextVariable | FormattedTextVariable | ImageVariable | ListVariable | GroupVariable;
+    type Variable = ShortTextVariable | LongTextVariable | FormattedTextVariable | ImageVariable | ListVariable | BooleanVariable | GroupVariable;
 
     enum VariableType {
         shortText = 'shortText',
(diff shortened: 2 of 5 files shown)


src/types/VariableTypes.ts
```diff
@@ -33,21 +33,23 @@ export interface Variable {
     isVisible: boolean;
     isReadonly: boolean;
     isRequired: boolean;
+    occurrences: number;
 }
 
 export interface ImageVariable extends Variable {
-    occurrences: number;
     src?: ImageVariableSource;
```
olegsolodukhin, line 40: @NicolaVerbeeck Just noticed, that variable `value` field is called `src` for ImageVariable and `selected` for list variable. I think it is better to follow the same structure and call all of them `value`, isn't it?

types/Actions.d.ts
```diff
@@ -167,6 +168,11 @@ declare module 'grafx-studio-actions' {
         selected?: string;
     }
 
+    export interface BooleanVariable extends BaseVariable {
+        type: VariableType.boolean;
+        value: boolean;
+    }
+
     interface GroupVariable extends BaseVariable {
         type: VariableType.group;
```
olegsolodukhin, line 177: Sorry, just noticed, but why there is no `value` for this variable type? If it is present for all other types.

olegsolodukhin (approved):

types/Actions.d.ts
Dvergar: Indeed, group is just a holder variable to group children. It has no value, children refer to the group by its id.

src/types/VariableTypes.ts
Dvergar: I agree it's not consistent. And I don't know if we want to change this now. It means migration > breaking change > WRS has to update 🥲

 For the list, though, what would be the value: items or selected?

Dvergar: @Matthiee @NicolaVerbeeck @evan-mcgeek @olegsolodukhin 

Not in scope of this ticket but can I have your input on this:

Maybe indeed `ImageVariable.src` should be `ImageVariable.value` & `ListVariable.items/selected` should be `ImageVariable.value` to be consistent with the other variables.

If we agree, two thoughts comes to my mind:
- It's enough w/ the breaking-changes for now
- If we are making breaking-changes, it's better to have all of them condensed in the same time frame

WDYT?

NicolaVerbeeck: ListVariable.value would then be the new name for ListVariable.selected?

NicolaVerbeeck (approved):

evan-mcgeek (approved):

src/types/VariableTypes.ts
olegsolodukhin: for the list - value is selected. Similarity: boolean has two options - true/false, the value is the selected one; String - all possible combinations of letters (possible), the value is the "selected" one; for the list we define possible values and value refers to the chosen one.

srares76 (approved):

PR #316, opened 2023-07-10
[Feature] Remove url variable source

Dvergar:
This PR removes the url variable source and removes the SDK method `setSource` in the variable controller.

* All variable image assets are going through a `media-url` embedded connector.
* `MediaConnectorImageVariableSource` is renamed `ConnectorImageVariableSource`.
* `UrlImageVariableSource` and `ImageVariableSourceType` are removed.
* SDK methods have not changed but the types have (`setSource` is removed).

## Related tickets

-   https://chilipublishintranet.atlassian.net/browse/EDT-991

Can be tested against https://studio.chiligrafx-dev.com/?workspace=latest&engine=314


src/controllers/VariableController.ts
```diff
@@ -235,10 +233,10 @@ export class VariableController {
      * @param id the id of the variable to update
      * @param source the new variable source
      */
-    setSource = async (id: string, source?: ImageVariableSource) => {
+    setSource = async (id: string, source?: ConnectorImageVariableSource) => {
```
Matthiee, line 236:
```suggestion
```

`setSource` is marked as deprecated. The only reason it is still here is because `UrlImageVariableSource` was still there. Now that it will be removed from the engine, I believe this method can be completely removed. 

We already had at least one SDK release where the deprecation was added so should be fine to remove it? 

Thoughts? @Dvergar @brapoprod

Dvergar: I had this thought about maybe giving integrators the time to adapt.

But If they need to refactor their code anyway (new image var source type) I would vote to remove it altogether.

pkgacek (dismissed):

src/controllers/VariableController.ts
Dvergar: _Resolved as Brian approved the removal via chat._

Matthiee (dismissed):

srares76 (approved):

pkgacek:
Any news on this one?

pkgacek (approved):

Matthiee:
> @abdelhalimkhouas had some concerns about WRS getting broken after this merge, the breakage should be handled by
> 
> https://chilipublishintranet.atlassian.net/browse/WRS-1262 https://chilipublishintranet.atlassian.net/browse/WRS-1263
> 
> Technically, yes, WRS might break when we are manipulating variable w/ connector sources as connectorId is hardcoded to grafx-media on WRS. Some issues may be handled "ok" w/ some unexpected states but I wouldn't count on this.
> 
> I'll let the team handle the merge (as I'm going on holiday).

@pkgacek  We wanted to make sure these two are ready first, just to avoid any potential issues.

abdelhalimkhouas:
I added an export of `VariableType` and adjusted the `ImageFrame` to the new naming, further cleanup of types and enums in the SDK should happen f.e `ImageFrameConnectorSource` and `ImageSourceTypeEnum`

Matthiee (approved):

brapoprod (approved):

PR #325, opened 2023-08-02
[Feature] Update document & keep connectors

Dvergar:
Can be tested against https://studio.chiligrafx-dev.com/?workspace=latest&engine=356

We can now keep connector states betwen document loads.

## Related tickets

https://chilipublishintranet.atlassian.net/browse/EDT-933


pkgacek (approved):

Matthiee (approved):

brapoprod (approved):

Matthiee (approved):

src/types/DocumentTypes.ts
```diff
@@ -198,3 +198,7 @@ export type TemplateDocumentProperties = DocumentProperties;
 export interface ProjectDocumentProperties extends DocumentProperties {
     templateId: string;
 }
+
+export type DocumentLoadOptions = {
+    keepConnectors: boolean;
```
NicolaVerbeeck, line 203: Should we add the docs describing each option in here too?

NicolaVerbeeck (dismissed):

src/types/DocumentTypes.ts
brapoprod, line 203: should be optional

Dvergar: @brapoprod WDYT Brian?

It's not done for the other ones and it's still available in `load` doc atm.

brapoprod (approved):

src/types/DocumentTypes.ts
brapoprod: I indeed think it makes most sense on top of the load function

srares76 (approved):

NicolaVerbeeck (approved):

PR #327, opened 2023-08-08
[Feature] Visibility naming

Dvergar:
## Related tickets

- https://chilipublishintranet.atlassian.net/browse/EDT-1076

## Breaking changes

- Document frame visibility property is renamed to `isVisible` from the old `included` (handled by a migration)
- Actions now set frame visibility with `setVisible` intead of `include`
- Actions now get frame visibility state with `isVisible` instead of `included`

## New SDK method/deprecation
- SDK default method for setting frame visibility is `setIsVisible` instead of `setVisibility`
- `setVisibility` still works but is deprecated


srares76 (approved):

pkgacek (approved):

Dvergar:
Renamed title from fix to feature as I understand this is needed for breaking changes & auto-versioning.

PR #357, opened 2023-09-27
[Feature] Error handling by throw

brapoprod:
This PR is a POC on throwing errors.
The SDK will now throw errors when something goes wrong instead of success status false.

### DIFF
--- a/packages/sdk/src/utils/EditorResponseData.ts
+++ b/packages/sdk/src/utils/EditorResponseData.ts
@@ -3,7 +3,12 @@ import type { EditorResponse } from '../types/CommonTypes';
 export function getEditorResponseData<T>(response: EditorResponse<unknown>, parse = true): EditorResponse<T> {
     try {
         if (!response.success) {
-            throw new Error(response.error ?? 'Yikes, something went wrong');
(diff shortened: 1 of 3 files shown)


pkgacek (approved):

packages/sdk/src/utils/EditorResponseData.ts
```diff
@@ -1,17 +1,24 @@
...
-        parsedData:
-            response.success && response.data
-                ? parse
-                    ? (JSON.parse(response.data as string) as T)
-                    : (response.data as unknown as T)
-                : null,
-    };
+    try {
+        if (!response.success) {
+            throw new Error(response.error ?? 'Yikes, something went wrong');
```
Matthiee, line 6: Can this throw a `SDKError` that has both `error`, `status` fields from `EditorResponse`?

brapoprod: I rather wouldn't extend the error object so I made a small modificaiton to the return, can you have a look if this suits your needs?

NicolaVerbeeck (approved):

pkgacek (approved):

PR #361, opened 2023-10-05
[Feature] Add width & height to media detail model

Dvergar:
This PR adds `width` & `height` to the model returned by the connector `detail` method. `MediaDetail` model is now diverging from `Media` (query method).

## Related tickets

https://chilipublishintranet.atlassian.net/browse/EDT-1054

### DIFF
--- a/packages/sdk/src/controllers/MediaConnectorController.ts
+++ b/packages/sdk/src/controllers/MediaConnectorController.ts
@@ -8,7 +8,7 @@ import {
     QueryPage,
 } from '../types/ConnectorTypes';
 import { CallSender } from 'penpal';
-import { Media, MediaDownloadType } from '../types/MediaConnectorTypes';
+import { Media, MediaDetail, MediaDownloadType } from '../types/MediaConnectorTypes';
 
 /**
  * The MediaConnectorController is responsible for all communication regarding media connectors.
@@ -65,7 +65,7 @@ export class MediaConnectorController {
         const res = await this.#editorAPI;
         return res
             .mediaConnectorDetail(id, mediaId, JSON.stringify(context))
-            .then((result) => getEditorResponseData<Media>(result));
+            .then((result) => getEditorResponseData<MediaDetail>(result));
     };
 
     /**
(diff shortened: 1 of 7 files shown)


pietervp:
Also update https://github.com/chili-publish/studio-sdk/blob/main/types/MediaConnector.d.ts to reflect the same changes, would be good to review if the complete api surfacce still matches what we expect.

pietervp (changes requested):
Also update https://github.com/chili-publish/studio-sdk/blob/main/types/MediaConnector.d.ts to reflect the same changes, would be good to review if the complete api surfacce still matches what we expect.

Dvergar:
> Also update https://github.com/chili-publish/studio-sdk/blob/main/types/MediaConnector.d.ts to reflect the same changes, would be good to review if the complete api surfacce still matches what we expect.

Ooops, you mentioned it in chat and i completely overlooked that.

So I updated both media & font connectors, we do not have preview (font) in getcapabilities so this probably needs to be added as well but i'll prefer not to do it in this ticket.

Probably some more cleanup related to that will be done in scope of >> https://chilipublishintranet.atlassian.net/browse/EDT-1172

@evan-mcgeek May I have your eyes 👀 on the font connector side?

PS: I'll split again MediaCapabilities/FontCapabilities

Dvergar:
Just adding some findings.

We do have 

```typescript
export enum MediaDownloadType {
    LowResolutionWeb = 'lowresWeb',
    HighResolutionWeb = 'highresWeb',
}
```

Was it an intended constraint to omit `outputVideo` & `outputPdf`?
Is it a standard for you WRS to start an enum entry with an uppercase?

evan-mcgeek (approved):
LGTM! The only thing left is to add a `preview` capability to the font connector

brapoprod (approved):
seems non-breaking to me

pietervp:
> Just adding some findings.
> 
> We do have
> 
> ```ts
> export enum MediaDownloadType {
>     LowResolutionWeb = 'lowresWeb',
>     HighResolutionWeb = 'highresWeb',
> }
> ```
> 
> Was it an intended constraint to omit `outputVideo` & `outputPdf`? Is it a standard for you WRS to start an enum entry with an uppercase?

That was intentional, because WRS should only ever use one of those.

pietervp (approved):

PR #367, opened 2023-10-10
[Fix] Note on eventual consistency

Dvergar:
Documentation on eventual consistency.

### DIFF
--- a/packages/sdk/src/controllers/SubscriberController.ts
+++ b/packages/sdk/src/controllers/SubscriberController.ts
@@ -303,6 +303,12 @@ export class SubscriberController {
      *
      * stateChanged - this event will be triggered by connector states: loading, loaded, running, ready, error
      * @param connectorEvent Stringified object of ConnectorEvent
+     * 
+     * **Eventual Consistency**
+     *
+     * Because of the nature of an async stream, be aware that some behaviors such
+     * as undo & redo might create scenarios where you're reacting on an event tied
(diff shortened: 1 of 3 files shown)


brapoprod (dismissed):

packages/sdk/src/controllers/SubscriberController.ts
```diff
@@ -217,6 +217,12 @@ export class SubscriberController {
      *
      * stateChanged - this event will be triggered by connector states: loading, loaded, running, ready, error
      * @param connectorEvent Stringified object of ConnectorEvent
+     * 
+     * **Eventual Consistency**
+     *
+     * Because of the nature of an async stream, be aware that some behaviors such
+     * as undo & redo might create scenarios where you're reacting on an event tied
+     * to a connector that has an old state.
```
Matthiee, line 225: This applies to every event that is being sent from the engine. So maybe there is a better place to put this @brapoprod ?

Matthiee (dismissed):

pkgacek (approved):

packages/sdk/src/controllers/SubscriberController.ts
brapoprod: maybe should be added to the readme so that it is explained on the SDK homepage (both github as in src)

Matthiee (approved):

psamusev (approved):

PR #371, opened 2023-10-11
[EDT-1186] Change downloadtypes according to new engine rules

pietervp:
DownloadType / DownloadIntent | *web* (= browser based editing session) | *print* (= PDF output, High res image output) | *animation* (= Gif, mp4 output, compression will be applied)
-- | -- | -- | --
*thumbnail* | ✓ (e.g., Displaying small preview images on a webpage) |   |  
*mediumres* | ✓ (e.g., Displaying larger images on a webpage without slowing down load times) | ✓ (e.g., Printing a decent-quality image where high resolution is not crucial) |  
*highres* | ✓ (e.g., Downloading a high-quality image for use on a high-resolution display) | ✓ (e.g., Printing a high-quality, large-scale image) | ✓ (e.g., Using as a frame in a high-quality animation)
*fullres* (PDF / PNG / JPEG) Hints to the fact its the file other formats are based on (media.pdf for illustrator, indesign, svg, etc files) Or just the original file when we are talking about most images |   | ✓ (e.g., Printing the original PDF or image file in its highest quality) |  
*original* | ✓ | ✓ | ✓

DownloadType: `[ Thumbnail, MediumRes, HighRes, FullRes, Original ]`
DonwloadIntent: `[Web, Print, Animation ]`

DownloadIntent is an optional parameter which gives the Connector developer a clue about what the download will be used for. IS for advances use cases only.

Env API ticket to be able to provide vector files for GraFx Media: https://chilipublishintranet.atlassian.net/browse/GRAFX-2490


brapoprod (approved):
lgtm but wondering why we deprecate something here?
since we're introducing breaking changes we could just maybe remove it and have no parseDeprecatedMediaDownloadType? What is the riskt?

abdelhalimkhouas:
I agree with @brapoprod, I assumed that these PR's https://github.com/chili-publish/studio-workspace/pull/716/files and https://github.com/chili-publish/studio-ui/pull/94 will habdle using the new types

abdelhalimkhouas (approved):

pietervp:
> lgtm but wondering why we deprecate something here? since we're introducing breaking changes we could just maybe remove it and have no parseDeprecatedMediaDownloadType? What is the riskt?

I thought it might be a good idea, but on second thought it's indeed maybe overkill. I will fix it in a follow up PR

PR #377, opened 2023-10-17
[PRODUCT-260] Move connector types to separate package

pietervp:
This PR moves the connector related type definition files to a separate package in order to:

* Release the package individually to npm
* Consume this package by connector developers
* Get a version of connectors definition usable by nodejs (connector test framework)


srares76 (approved):

packages/connectors/src/Connector.Shared.ts
```diff
@@ -0,0 +1,77 @@
...
+     */
+    headers?: Dictionary;
+    /**
+     * A cryptographic hash of the resource to be fetched by request. Sets request's integrity.
+     */
+    integrity?: string;
+    /**
+     * A string to set request's method.
+     */
+    method?: string;
```
brapoprod, line 40: could be probably an enum or string literal?

brapoprod (approved):

packages/connectors/src/Connector.Shared.ts
pietervp: Will address this in follow up PR's. This initial PR is just moving whatever we have to a new location. Created https://chilipublishintranet.atlassian.net/browse/EDT-1248 for this

PR #382, opened 2023-10-26
[FIX] Document PR title flow

NicolaVerbeeck:


srares76 (approved):

brapoprod (approved):

brapoprod:
> LGTM, only remark is that now this behavior is not enforced. In the sense that I could choose to have a PR without any of the predefined tags ([FIX], etc). In this case it is automatically treated as a patch version. Which I believe is fine, but enforcing it could make people think about it more explicitly

can be an increment over time no?

images/pr_flow.svg
Matthiee, line 1: So we can never have an NPM version with only the patch version being increased?

Matthiee (approved):

images/pr_flow.svg
NicolaVerbeeck: @brapoprod did we miss a case? :)

Dvergar (approved):

images/pr_flow.svg
brapoprod: yes we can

PR #386, opened 2023-11-06
[Feature] Add clipboard controller

Dvergar:
This PR adds a `ClipboardController` to copy/paste/cut frames and a new `duplicateFrames` in the `FrameController`.

The clipboard controller is opiniated, it does only support frames as of now; handling the clipboard regardless of the context frame/text brought up several issues on TLE and some focus edge-cases.

Right now the copy/paste operations handled by the shorcuts from the editor are entirely driven by the engine.

For integrators we have 4 clipboard related methods:
- `copyFrames`
- `cutFrames`
- `pasteFrames`
- `getContentType`

The management of the clipboard (the access to the OS clipboard) in this case is handled by the SDK.

The reason is that the clipboard only works when the window is focused. If you have the focus on a side panel, then you lose the focus to the engine, hence the access to the clipboard (handled by the engine).

Having the SDK directly accessing the clipboard allows for integrators to add buttons & context menus for a e.g. a copy operation without having to manually deal with focus management.

## PR Guidelines

- [x] I have read the [contribution and PR guidelines](CONTRIBUTING.md)

## Related tickets

-   [EDT-911](https://chilipublishintranet.atlassian.net/browse/EDT-911)

[EDT-911]: https://chilipublishintranet.atlassian.net/browse/EDT-911?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sdk/src/types/SelectionTypes.ts
Dvergar, line 1: Note to myself: to remove.

packages/sdk/src/controllers/ClipboardController.ts
```diff
@@ -0,0 +1,67 @@
...
+    };
+
+    /**
+     * This method will cut the frames to the OS clipboard as a json frame
+     * @param ids An array of all frame ids you want to cut
+     * @returns
+     */
+    cutFrames = async (ids: Id[]) => {
+        const res = await this.#editorAPI;
+        const frameData = await res.cutFrames(ids);
```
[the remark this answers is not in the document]
Dvergar: Yes.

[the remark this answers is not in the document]
Dvergar: (if that's not so explicit I'll add it in the code doc)

packages/sdk/src/controllers/ClipboardController.ts
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        return getEditorResponseData<null>(frameData);
+    };
+
+    /**
+     * This method will paste the OS clipboard content to the editor as frames
+     * @returns
+     */
+    pasteFrames = async () => {
```
brapoprod, line 57: shouldn't this have an optional parameter (or override) that accepts 1 frame ID? then the frames will be pasted below or above (product question) the used frame

Dvergar: Following a call, we're going to paste new frames on top of the used frames by default & implementation will be on engine-side.

Matthiee (approved):

brapoprod (approved):

PR #389, opened 2023-11-13
[Feature] Introduce new subscription for multiple selected frames

evan-mcgeek:
This PR

## PR Guidelines

- [X] I have read the [contribution and PR guidelines](CONTRIBUTING.md)

## Related tickets

-   [EDT-1181](https://chilipublishintranet.atlassian.net/browse/EDT-1181)

[EDT-1181]: https://chilipublishintranet.atlassian.net/browse/EDT-1181?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sdk/src/controllers/SubscriberController.ts
```diff
@@ -1,6 +1,6 @@
-import { ConfigType, Id } from '../types/CommonTypes';
-import { MeasurementUnit } from '../types/LayoutTypes';
-import { ToolType } from '../utils/enums';
+import {ConfigType, Id} from '../types/CommonTypes';
+import {MeasurementUnit} from '../types/LayoutTypes';
+import {ToolType} from '../utils/enums';
```
evan-mcgeek, line 3: guys, is there something wrong with my prettier?

packages/sdk/src/controllers/SubscriberController.ts
```diff
@@ -66,21 +66,44 @@ export class SubscriberController {
...
+        const callBack = this.config.onSelectedFramesLayoutChanged;
+        const frames = JSON.parse(frameLayout);
+        callBack && callBack(frames[0]);
+    };
+
+    /**
+     * Listener on the state of the currently selected frames, if this changes, this listener will get triggered with the updates
+     * @param frameLayout Stringified array of FrameLayoutType objects
+     */
+    onSelectedFramesLayoutChanged = (frameLayout: string) => {
```
NicolaVerbeeck, line 85: nit: naming of the string parameter suggests single layout

packages/sdk/src/controllers/SubscriberController.ts
```diff
@@ -66,21 +66,44 @@ export class SubscriberController {
...
+        const frames = JSON.parse(frameContent);
+        callBack && callBack(frames[0]);
+    };
+
+    /**
+     * Listener on the state of the currently selected frame, it contains some basic information on the type of frame it is
+     * @param frameContent Stringified array of Frame objects
+     */
+    onSelectedFramesContentChanged = (frameContent: string) => {
+        const callBack = this.config.onSelectedFramesContentChanged;
```
NicolaVerbeeck, line 106: Nit naming of the string parameter

packages/sdk/src/controllers/SubscriberController.ts
```diff
@@ -66,21 +66,44 @@ export class SubscriberController {
         callBack && callBack(unit as MeasurementUnit);
     }
 
+
     /**
-     * Listener on the state of the currently selected frame, if this changes, this listener will get triggered with the updates
-     * @param frameLayout Stringified object of FrameLayoutType
+     * @deprecated use `onSelectedFramesLayoutChanged` instead
+     * Listener on the state of the currently selected frames, if this changes, this listener will get triggered with the updates
+     * @param frameLayout Stringified array of FrameLayoutType objects
```
NicolaVerbeeck, line 73: Update the comment to reflect that it will be the first frame when multiple are selected

packages/sdk/src/controllers/SubscriberController.ts
```diff
@@ -66,21 +66,44 @@ export class SubscriberController {
...
         callBack && callBack(JSON.parse(frameLayout));
     };
 
     /**
+     * @deprecated use `onSelectedFramesContentChanged` instead
      * Listener on the state of the currently selected frame, it contains some basic information on the type of frame it is
-     * @param frameContent Stringified object of FrameType
+     * @param frameContent Stringified array of Frame objects
      */
     onSelectedFrameContentChanged = (frameContent: string) => {
```
NicolaVerbeeck, line 95: Update the comment to reflect that it will be the first frame when multiple are selected

packages/sdk/src/tests/controllers/SubscriberContoller.test.ts
```diff
@@ -93,15 +97,15 @@ describe('SubscriberController', () => {
         expect(mockEditorApi.onScrubberPositionChanged).toHaveBeenCalledTimes(1);
         expect(mockEditorApi.onScrubberPositionChanged).toHaveBeenLastCalledWith('test');
     });
-    it('Should be possible to subscribe to onSelectedFrameLayoutChanged', async () => {
-        await mockedSubscriberController.onSelectedFrameLayoutChanged('2');
-        expect(mockEditorApi.onSelectedFrameLayoutChanged).toHaveBeenCalledTimes(1);
-        expect(mockEditorApi.onSelectedFrameLayoutChanged).toHaveBeenCalledWith(2);
+    it('Should be possible to subscribe to onSelectedFramesLayoutChanged', async () => {
```
NicolaVerbeeck, line 100: Re-add the tests for the deprecated methods to ensure they produce the intended results

NicolaVerbeeck (changes requested):

NicolaVerbeeck (approved):

Dvergar (approved):

packages/sdk/src/controllers/SubscriberController.ts
```diff
@@ -66,22 +66,47 @@ export class SubscriberController {
...
+     * Whenever multiple frames are selected, it always will be the first selected `FrameLayoutType`
+     * @param frameLayout Stringified array of `FrameLayoutType` objects
      */
     onSelectedFrameLayoutChanged = (frameLayout: string) => {
-        const callBack = this.config.onSelectedFrameLayoutChanged;
-        callBack && callBack(JSON.parse(frameLayout));
+        const callBack = this.config.onSelectedFramesLayoutChanged;
+        const frames = JSON.parse(frameLayout);
+        callBack && callBack(frames[0]);
+    };
```
brapoprod, line 80: why deprecate it? it is still a valid call

packages/sdk/src/controllers/SubscriberController.ts
```diff
@@ -66,22 +66,47 @@ export class SubscriberController {
...
+     * Listener on the state of the currently selected frames, if this changes, this listener will get triggered with the updates
+     * @param framesLayout Stringified array of FrameLayoutType objects
+     */
+    onSelectedFramesLayoutChanged = (framesLayout: string) => {
+        const callBack = this.config.onSelectedFramesLayoutChanged;
+        callBack && callBack(JSON.parse(framesLayout));
     };
 
     /**
+     * @deprecated use `onSelectedFramesContentChanged` instead
```
brapoprod, line 92: same here, call works and uses the newest endpoint so I wouldn't deprecate it

packages/sdk/src/controllers/SubscriberController.ts
Matthiee: Indeed looks like something is wrong here, @brapoprod any idea?

packages/sdk/src/controllers/SubscriberController.ts
```diff
@@ -66,22 +66,47 @@ export class SubscriberController {
...
+     * Listener on the state of the currently selected frames
+     * Whenever multiple frames are selected, it always will be the first selected `FrameLayoutType`
+     * @param frameLayout Stringified array of `FrameLayoutType` objects
      */
     onSelectedFrameLayoutChanged = (frameLayout: string) => {
-        const callBack = this.config.onSelectedFrameLayoutChanged;
-        callBack && callBack(JSON.parse(frameLayout));
+        const callBack = this.config.onSelectedFramesLayoutChanged;
+        const frames = JSON.parse(frameLayout);
+        callBack && callBack(frames[0]);
```
Matthiee, line 78: In `sdk.ts` this is added to the subscriber methods and made available to the engine as a callback. However, this doesn't exist on the engine anymore, to me it seems that this won't work. 

This code should be moved to `onSelectedFramesLayoutChanged`, that one should invoke both callbacks instead of having the callback here. 

Maybe I'm wrong here, haven't tried linking the updated SDK to the WRS

Matthiee (changes requested):

packages/sdk/src/controllers/SubscriberController.ts
evan-mcgeek: I got it, the path Prettier package has changed since the project has been restructured

packages/sdk/src/controllers/SubscriberController.ts
evan-mcgeek: you are right, it didn't work. a better approach has been pushed

packages/sdk/src/controllers/SubscriberController.ts
evan-mcgeek: The engine-side update now exposes an array of frames instead of just one. I think it's best to avoid embedding specific display logic within the SDK. Instead, giving integrators the choice to decide how they want to present this data — whether focusing on a single frame's properties or not.

packages/sdk/src/controllers/SubscriberController.ts
evan-mcgeek: same answer as above

Matthiee (dismissed):

Matthiee (approved):

Dvergar (approved):

abdelhalimkhouas (approved):

packages/sdk/src/controllers/SubscriberController.ts
brapoprod: My question is not persé why you deprecate it, I know why, but I don't feel like this is necessary.
The behaviour of the call, still stays the same so the function is still a valid one imo. Since we use the new call with index 0, this is perfect behaviour and one breaking change less to worry about?

evan-mcgeek: the behavior of the call remains the same since there is a workaround for that (to avoid breaking changes in the first place). I don't think that is a correct approach and still convinced that it shouldn't be there. No strong opinion though

brapoprod: yes you might be right, I'll resolve convos again

brapoprod (approved):

PR #415, opened 2024-01-26
[Feature] Add view mode

Dvergar:
This PR adds view mode.

## PR Guidelines

## Related tickets

https://chilipublishintranet.atlassian.net/browse/WRS-1552
https://chilipublishintranet.atlassian.net/browse/WRS-1551
https://chilipublishintranet.atlassian.net/browse/EDT-1293
https://chilipublishintranet.atlassian.net/browse/PRODUCT-388


srares76 (approved):

Matthiee (approved):

NicolaVerbeeck (approved):

Matthiee (approved):

packages/sdk/src/controllers/ViewModeController.ts
```diff
@@ -0,0 +1,30 @@
...
+    constructor(editorAPI: EditorAPI) {
+        this.#editorAPI = editorAPI;
+    }
+
+    /**
+     * This method sets a new view mode
+     * @param viewMode view mode of the editor
+     * @returns
+     */
+    set = async (viewMode: ViewMode) => {
```
brapoprod, line 26:
```suggestion
    setViewMode = async (viewMode: ViewMode) => {
```
To be inline with other controllers. You don't know if we're going to extend this

Dvergar: I wanted to ask on Studs as well because indeed `ToolController` is the same as you showed.

What is "_extending_" in this context? Won't `viewMode.set` always have the same intent?

brapoprod (changes requested):
add controller to documentation

packages/sdk/src/controllers/ViewModeController.ts
brapoprod: you are right, but question is if the controller isn't too specific?

srares76 (approved):

packages/sdk/src/types/ViewModeTypes.ts
```diff
@@ -0,0 +1,4 @@
+export enum ViewMode {
```
abdelhalimkhouas, line 1: Can u plz export this so it can be used in studio workspace and other integrations

abdelhalimkhouas (approved):

abdelhalimkhouas (approved):

PR #457, opened 2024-05-06
[Fix] Consolidate connector shared API

Dvergar:
I moved common connector code between `sdk` & `connector` packages to the `connector` package via the `external` folder, does that impact packaging?
I did not found a way to properly resolve the path of `@connectorShared`, I tried different resolvers & lint parameters, it sometimes solved it and then just reappeared. Some help from js experts are needed here.

This shouldn't be a breaking-change.

There were some wrong types, or wrongly-typed properties.

Generally an enum on the SDK becomes a string literal type on the connector side, should we streamline this? (breaking)

## PR Guidelines

## Related tickets

https://chilipublishintranet.atlassian.net/browse/EDT-1248


packages/sdk/src/types/ConnectorTypes.ts
```diff
@@ -1,3 +1,4 @@
+import { ConnectorConfigValue } from '@connectorShared/Connector.Shared.external';
```
Matthiee, line 1: Can't you add the `connector` package in `packages.json` as a dependency?

Dvergar: Yes but the gist of what I got from Pieter was that we did not want to overshare connector to SDK since the whole request & co part is unneeded. Maybe for autocompletion? Summoning @pietervp

psamusev: can we move this external things to separate package and mark it as private? and configure web pack to include its content into others public packages?

Dvergar: Fine by me if the team agrees, @brapoprod @abdelhalimkhouas ?

abdelhalimkhouas: Sounds good to me

brapoprod: I don't like the seperation of this code, that defeats the purpose of having a monorepo, we should bundle it inside the SDK bundle, not seperate, or am i missing the point here?

Dvergar: I don't know the details on the how to do that. The only goal for me is to be able to access the same connector models from both the `connector` package and the `sdk` package as it is too easy to diverge without noticing that one is wrong or more up to date than the other - same - model.

Matthiee: > Can't you add the connectors package in package.json as a dependency?

I think what @pietervp means is to not combine both packages, as in that case, I agree it doesn't make sense. 

However, what I'm suggesting is a bit different and I don't see the reason being an actual problem. Even if they are bundled, things like tree shaking still exist so not sure if it is an actual issue. 

Also not sure if my suggestion would make things easier compared to the other suggestions.

Dvergar: @Matthiee We got a discussion before your comment. We dropped the package approach (no strong opinion on this).

@brapoprod @abdelhalimkhouas @alexandraFlavia9 I reverted the changes, merged the types but still have some strange module resolution. Fell free to take over the PR.

Dvergar:
Closed in favor of https://github.com/chili-publish/studio-sdk/pull/612

PR #462, opened 2024-05-27
[Feature] [EDT-1482] Add `NumberVariable` to Actions

Matthiee:
This PR adds `NumberVariable` support to `Actions` and helper methods.

## PR Guidelines

## Related tickets

-   [EDT-1482](https://chilipublishintranet.atlassian.net/browse/EDT-1482)

[EDT-1482]: https://chilipublishintranet.atlassian.net/browse/EDT-1482?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/actions/scripts/prepare-release.mjs
+++ b/packages/actions/scripts/prepare-release.mjs
@@ -14,6 +14,9 @@ async function main() {
     const packageJson = JSON.parse(await fs.readFile('package.json', 'utf-8'));
     const apiVersion = packageJson.apiVersion;
     const sdkVersion = packageJson.version;
+    const prId = process.env.CI_PR_ID;
+
+    const isPR = prId !== undefined;
 
     // Read studio-engine.json
     const studioEngineJson = JSON.parse(await fs.readFile('../sdk/editor-engine.json', 'utf-8'));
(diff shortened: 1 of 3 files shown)


packages/actions/scripts/prepare-release.mjs
```diff
@@ -14,6 +14,9 @@ async function main() {
     const packageJson = JSON.parse(await fs.readFile('package.json', 'utf-8'));
     const apiVersion = packageJson.apiVersion;
     const sdkVersion = packageJson.version;
+    const prId = process.env.CI_PR_ID;
```
Dvergar, line 17: What is this for?

packages/actions/types/Actions.d.ts
```diff
@@ -495,6 +552,11 @@ declare module 'grafx-studio-actions' {
              */
             getBooleanValue(name: string | Variable): boolean;
 
+            /**
+             * Gets the number value of a variable
+             */
+            getNumberValue(name: string | Variable): number;
```
Dvergar, line 558: I'm a bit lost with all the helpers, don't we want a `getNumber`(variable) here as well?

Matthiee: In this file only the typings are defined, these should match with the `studio` object being exposed by the `engine`. 

> don't we want a getNumber(variable) here as well?

I don't think so, `studio.variables.byName("myNumberVar")` already returns a `NumberVariable` instance. But it is typed as the generic `Variable`. 

To help with this an actual helper function is defined here:

https://github.com/chili-publish/studio-sdk/pull/462/files#diff-2e9f181658fcc70ebad720fcb2c98c9a0a64a357b9e609c73368f64a6de52a16R44

packages/actions/scripts/prepare-release.mjs
Matthiee: To have these files available on a per PR basis. Right now these files are overwritten on each SDK PR build 😕

Dvergar (approved):

packages/actions/scripts/prepare-release.mjs
Matthiee, line 17: @brapoprod this script runs on every PR build. 

The problem here is that I updated the files, but if another PR targeting the same SDK version is created/updated a new build will happen and overwrite my version (modified) of the actions with their version (unmodified). 

As a workaround for now I added this, so that the files are added in `.../prs/462/...` as well, so at least QA and I have a stable reference to my version of the files. 

Any thoughts?

brapoprod (approved):

PR #476, opened 2024-06-25
[FEATURE] Strict typings on connectors mapping

psamusev:
This PR adds type safety when working with ConnectorMapping. It also adds possibility to retrieve specified mapping for provided direction (with again type safety around `value` field)

## PR Guidelines

### DIFF
--- a/packages/sdk/src/controllers/ConnectorController.ts
+++ b/packages/sdk/src/controllers/ConnectorController.ts
@@ -2,7 +2,6 @@ import { ConnectorOptions, EditorAPI, EditorResponse, Id } from '../types/Common
 import {
     ConnectorState,
     ConnectorStateType,
-    ConnectorMapping,
     ConnectorMappingType,
     ConnectorRegistration,
     ConnectorInstance,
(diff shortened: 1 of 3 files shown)


Matthiee (approved):

packages/sdk/src/controllers/ConnectorController.ts
```diff
@@ -2,10 +2,13 @@ import { ConnectorOptions, EditorAPI, EditorResponse, Id } from '../types/Common
 import {
     ConnectorState,
     ConnectorStateType,
-    ConnectorMapping,
+    ConnectorMappingType,
```
brapoprod, line 5: why renaming this? this is a breaking change...

psamusev: It's not renaming. I've added a new type that uses as return value of ConnectorController method. The previous class with name `ConnectorMapping` still exists

brapoprod: I see, my bad. It's weird UI from github :)

brapoprod (approved):

PR #480, opened 2024-07-08
[EDT-1549] set/resetPlaceholder methods

evan-mcgeek:
This PR adds `setPlaceholder` and `resetPlaceholder` methods for a `Variable`

## PR Guidelines

## Related tickets

-   [EDT-1549](https://chilipublishintranet.atlassian.net/browse/EDT-1549)

## Screenshots

[EDT-1549]: https://chilipublishintranet.atlassian.net/browse/EDT-1549?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sdk/src/types/VariableTypes.ts
```diff
@@ -43,6 +43,7 @@ export interface Variable {
     isReadonly: boolean;
     isRequired: boolean;
     occurrences: number;
+    placeholder?: string;
```
Dvergar, line 46: Not really a comment to this PR but I know we use optional as the default for nullables in the engine in a lot of those models.

The engine models that are not explictly set as `includeIfNull` will shorten the model json by removing those nullable properties.
In such situations, my understanding is that reaching `placeholder` will return `undefined`; but what about a scenario where the engine is explictly serializing `null`, I assume the model will break since we are in strict mode!?

Anyone to enlighten me?

Dvergar (approved):

packages/sdk/src/types/VariableTypes.ts
Dvergar: It has been working fine so far but I feel like we should slowly make the union type the default (we have a lot of cases) if only to be more explicit and aligned with engine models (it's probably flutter team's job though).

Matthiee (approved):

Dvergar (approved):

PR #482, opened 2024-07-11
GRAFX-3507 Support multiple event handlers

pietervp:
## Problem
When using the SDK, we expose a lot of event handlers to the integrator. For example, we can execute a callback whenever the selected frame changed. There is one big (not obvious) caveat, the handler can only be defined once. This means that each event can only be handled by 1 consumer. This is especially bad when using studio-ui, and you want to tap into some of these event yourself as integrator. Once you do that, you probably have broken the studio-ui interface by overriding it’s subscription.

## Proposal
* Deprecate all properties in SDK.config.on<eventName> and move logic to SDK.config.events where each property on<eventName> will no longer be a simple callback, but a class to manage subscriptions. 
* Remain backwards compatible, the events properties will by default use the legacy SDK.config.on<eventName> callback when the event is triggered. We also keep supporting overriding that handler. 
* Follow Up: Workspace and studio-ui should move to the events based approach when registering event handlers. (first we need SDK PR)

### Code Highlights
* The SDK object constructor remains intact (`ConfigType` does not include `events` property). Although the SDK constructor starts with wrapping this config object to a `RuntimeConfigType` which adds the events. https://github.com/chili-publish/studio-sdk/pull/482/files#diff-04689b788dad65b6ce4b43e7344d3b854b1a95ad2c8692c6ece28ccd7d9a96b4R93 
* `EventHelper.ensureSubscriptions` => this wraps the `ConfigType` to a `RuntimeConfigType`, automatically registering the legacy handlers already defined in `ConfigType`. https://github.com/chili-publish/studio-sdk/pull/482/files#diff-8cc56e41f48f0a52e82e4c279dba504747c37e1c76bf0d79b802505878eb56a7R69
* `EventSubscription` => controls the multiple callbacks (both registering and execution) https://github.com/chili-publish/studio-sdk/pull/482/files#diff-8cc56e41f48f0a52e82e4c279dba504747c37e1c76bf0d79b802505878eb56a7R23

### Open Questions
* Naming: do we keep `SDK.config.events` or are there better naming options?
* Naming: do we keep `SDK.config.handlers` to indicate callbacks that require to return a result


packages/sdk/src/utils/EventSubscription.ts
```diff
@@ -0,0 +1,165 @@
...
+    private callbacks: Map<string, T> = new Map();
+    legacyEventHandler: (() => T | undefined) | undefined;
+
+    constructor(legacyEventHandler?: () => T | undefined) {
+        super();
+        this.legacyEventHandler = legacyEventHandler;
+    }
+
+    // Register a new callback with an optional name
+    registerCallback(callback: T, name?: string): string {
```
psamusev, line 34: when do we might need an optional name useCase?  Wouldn't be simpler to subscribe to the SDK defined events only?

packages/sdk/src/utils/EventSubscription.ts
```diff
@@ -0,0 +1,166 @@
...
+    private generateUniqueName(): string {
+        return `callback_${Math.random().toString(36).substr(2, 9)}`;
+    }
+}
+
+export class EventHelper {
+    static ensureSubscriptions(config: ConfigType): RuntimeConfigType {
+        // we need a reference to the clone object in the eventsubscriptions, so we need to create the object in 2 steps
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
```
psamusev, line 73: a bit dangerous to use it, might bring some unpredictable mistakes

packages/sdk/src/controllers/SubscriberController.ts
```diff
@@ -115,14 +107,7 @@ export class SubscriberController {
      * @param authRefreshRequest Stringified object of AuthRefreshRequest
      */
     onAuthExpired = async (authRefreshRequest: string) => {
```
Dvergar, line 109: Maybe that one should be outside of the `subscriberController` concept as it does not follow the usual flow.

In this case, the engine is actually explictly **awaiting** for the handler on the integration to return something.
This is also why it's returning `null` by default: if no handler exists, then the engine will react on the returned `null` to unblock itself. If there is one, the integration will process the authentication and return a credentials object so that the engine can proceed to refetch the assets.

With the new refactoring there is none of this logic, there is probably a way to make it work but in the end that one subscriber (also with `onViewportRequested`) should actually exceptionally expect only one handle.

packages/sdk/src/controllers/SubscriberController.ts
```diff
@@ -135,45 +120,29 @@ export class SubscriberController {
      * zoom to page call.
      */
     onViewportRequested = () => {
```
Dvergar, line 122: Same comment as `onAuthExpired`.

Dvergar (changes requested):

packages/sdk/src/utils/EventSubscription.ts
pietervp: The name here is not used for the `eventName`, rather is just a key for the callback that was added. The main purpose is to be able to unsubscribe more easily. Imagine only needing to subscribe for a limited amount of time, for example. Maybe updating the parameter name to `key` would make that more clear?

packages/sdk/src/utils/EventSubscription.ts
pietervp: This was the best I could do, I know you TS skills are superior, feel free to take a shot at it. I added tests that cover all the essential parts, so if they stay green, you will be good

packages/sdk/src/controllers/SubscriberController.ts
pietervp: The logic is actually moved to https://github.com/chili-publish/studio-sdk/pull/482/files#diff-8cc56e41f48f0a52e82e4c279dba504747c37e1c76bf0d79b802505878eb56a7R46-R61. What I did was just executing all subscribers, and the return result of the latest subscriber will be used to return to engine. This still allows to override the `onViewportRequested` or `onAuthExpired`. I do agree it might be a bit obscure if we dont document this very well.

Alternatively we indeed need to make special cases for these where we only allow 1 subscription and do not allow overriding that. All in all I dont think we need to make the exception. People adding subscriptions to those methods probably know what they are doing.

Matthiee: All the "events" should have a return type of `void`. All the other ones are technically not "events", they are requests that require a singular handler. 

We should distinguish between them. 

----

Perhaps `clone.events` and `clone.requests` 🤔 

See: https://github.com/chili-publish/studio-sdk/pull/482/files#diff-8cc56e41f48f0a52e82e4c279dba504747c37e1c76bf0d79b802505878eb56a7R75-R149

pietervp: Good points.

* I introduced `handlers` next to `events`: https://github.com/chili-publish/studio-sdk/pull/482/files#diff-a5080035ef7a9f26c66428b08da906fd0d75994ea367df52cae7e305db2c33d4R36-R39
* They are managed by a `SingleEventSubscription` which does not allow adding additional handlers.

Dvergar: Thanks! One note on the new handlers section.

We used to have what could be a default handler that was returning `null`, the reasoning was to gracefully fail and/or to return a meaningful error message when an handler wasn't subscribed to.

Without those default handlers returning `null`, engine will await indefinitely until something is returned to it and will just hang without any hint.
e.g. on `main`, behavior is that if no viewport is returned, engine will catch the `null`, stop awaiting and return a message saying that the viewport handler needs to be defined.

Unless `trigger` can return `null` by default, or maybe just enforcing the handler to be defined or any other mechanism.

pietervp: I thought it was handled, so added an explicit test case, and you are right, it was returning undefined instead of null in this case. Should be good now!

packages/sdk/src/utils/EventSubscription.ts
```diff
@@ -0,0 +1,206 @@
...
+        let result;
+
+        if (this.legacyEventHandler) {
+            const callback = this.legacyEventHandler();
+            if (callback) {
+                result = callback(...args);
+            }
+        }
+
+        for (const callback of this.callbacks.values()) {
```
Matthiee, line 49: All of the event handlers are required to be sync? 
While the request handlers can be both async/sync?

Matthiee (commented):
I like this idea 💯 

I do have some concerns regarding the naming.

pietervp:
After receiving feedback from @Matthiee  and discussing offline with @brapoprod :

* Added a configuration option to register a logger in the SDK (https://github.com/chili-publish/studio-sdk/pull/482/files#diff-a5080035ef7a9f26c66428b08da906fd0d75994ea367df52cae7e305db2c33d4R33-R36)
  * This allows us to avoid `console.log` in the code, and gives integrators a way to get to errors / warning / info generated by us.
  * This is needed because of the async nature of the event callbacks, there is no good way for an integration to handle these exceptions by default.
* Every callback registration can define how errors during callback execution should be handled. (log / throw / remove) (https://github.com/chili-publish/studio-sdk/pull/482/files#diff-a5080035ef7a9f26c66428b08da906fd0d75994ea367df52cae7e305db2c33d4R57-R72 and https://github.com/chili-publish/studio-sdk/pull/482/files#diff-9b49133c3aadcf3d8efbb67c197a89e4e8009cc91ed27042842f342818dc2d41R44-R58)
  * This allows each subscription to define its own rules on how to handle errors during callback
* On the feedback about the registration keys, we think keeping it a string is cleaner and simpler then requiring an object which need to be tracked by the integrator. Strings can be easily managed on integration side, no leaks, etc
* Other naming changes were applied

packages/sdk/src/utils/EventSubscription.ts
```diff
@@ -0,0 +1,206 @@
...
+            result = callback(...args);
```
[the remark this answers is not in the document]
pietervp: Updated, see general comment

packages/sdk/src/utils/EventSubscription.ts
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    /**
+     * Registers a callback function with an optional key.
+     *
+     * @param callback - The callback function to register.
+     * @param key - An optional key to associate with the callback.
+     * @returns The generated or provided key associated with the callback.
+     */
+    registerCallback(callback: T, key?: string): string {
```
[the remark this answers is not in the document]
pietervp: Keeping it as string keeps things a bit easier on integration side (no danling references to object in SDK, transparent on how to use, etc). All things considered, changing this seems to add complexity more then it helps with anything.

```
const key : string = SDK.config.events.onXyz.subscribe(()={});
SDK.config.events.onXyz.unSubscribe(key);
```

packages/sdk/src/utils/EventSubscription.ts
```diff
@@ -0,0 +1,206 @@
...
+        const callbackName = key || this.generateUniqueName();
+        this.callbacks.set(callbackName, callback);
+        return callbackName;
+    }
+
+    /**
+     * Removes the callback associated with the specified name from the subscription.
+     * @param name - The name of the callback to remove.
+     */
+    unsubscribeCallback(name: string): void {
```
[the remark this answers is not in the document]
pietervp: See prev comment

packages/sdk/src/utils/EventSubscription.ts
```diff
@@ -0,0 +1,206 @@
...
+
+        for (const callback of this.callbacks.values()) {
+            result = callback(...args);
+        }
+
+        return result ?? null;
+    }
+}
+
+export class EventSubscription<T extends (...args: any[]) => any> extends SingleSubscription<T> {
```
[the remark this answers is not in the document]
pietervp: Done

packages/sdk/src/utils/EventSubscription.ts
```diff
@@ -0,0 +1,206 @@
...
+import { CornerRadiusUpdateModel } from '../types/ShapeTypes';
+import { SelectedTextStyle } from '../types/TextStyleTypes';
+import { Variable } from '../types/VariableTypes';
+import { ViewMode } from '../types/ViewModeTypes';
+import { Viewport } from '../types/ViewportTypes';
+import { ToolType } from './Enums';
+
+class EventSubscriptionBase {}
+
+export class SingleSubscription<T extends (...args: any[]) => any> extends EventSubscriptionBase {
```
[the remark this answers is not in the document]
pietervp: Done

packages/sdk/src/utils/EngineCallbackHandler.ts
```diff
@@ -0,0 +1,85 @@
...
+                    this.logger(LogLevel.error, LogCategory.event, `Error in callback ${key}: ${error}`);
+                }
+            } else if (errorBehavior === CallbackErrorBehavior.throw) {
+                throw error;
+            } else if (errorBehavior === CallbackErrorBehavior.remove) {
+                this.callbacks.delete(key);
+                if (this.logger) {
+                    this.logger(LogLevel.warn, LogCategory.event, `Removed callback ${key} due to error: ${error}`);
+                }
+            }
```
Matthiee, line 82: This is duplicated and could probably be extracted

Matthiee (approved):

packages/sdk/src/utils/EngineCallbackHandler.ts
psamusev: extracted

PR #504, opened 2024-08-14
[Feature] Add private data

Dvergar:
This PR

## PR Guidelines

## Related tickets

https://chilipublishintranet.atlassian.net/browse/EDT-1608

### DIFF
--- a/packages/sdk/src/controllers/VariableController.ts
+++ b/packages/sdk/src/controllers/VariableController.ts
@@ -9,6 +9,7 @@ import {
     Locale,
     NumberVariablePropertiesDeltaUpdate,
     PrefixSuffixDeltaUpdate,
+    PrivateData,
     Variable,
     VariableType,
 } from '../types/VariableTypes';
@@ -616,7 +617,7 @@ export class VariableController {
      */
     getPrivateData = async (id: string) => {
         const res = await this.#editorAPI;
-        return res.getVariablePrivateData(id).then((result) => getEditorResponseData<null>(result));
+        return res.getVariablePrivateData(id).then((result) => getEditorResponseData<PrivateData>(result));
     };
 
     private makeVariablesBackwardsCompatible(variables: Variable[]) {
--- a/packages/sdk/src/controllers/VariableController.ts
+++ b/packages/sdk/src/controllers/VariableController.ts
@@ -596,6 +596,29 @@ export class VariableController {
         return this.applyPrefixSuffixDeltaUpdate(id, update);
     };
 
+    /**
+     * Sets the private data for any variable
+     * @param id the id of the variable to update
+     * @param privateData the private data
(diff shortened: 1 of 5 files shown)


NicolaVerbeeck (approved):

packages/sdk/src/types/VariableTypes.ts
```diff
@@ -45,6 +45,7 @@ export interface Variable {
     occurrences: number;
     placeholder?: string | null;
     helpText?: string | null;
+    privateData: Record<string, string>;
```
Matthiee, line 48: Maybe we should have a private data alias?

packages/sdk/src/controllers/VariableController.ts
```diff
@@ -596,6 +596,29 @@ export class VariableController {
...
+    };
+
+    /**
+     * Gets the private data for any variable
+     * @param id the id of the variable
+     * @returns the private data
+     */
+    getPrivateData = async (id: string) => {
+        const res = await this.#editorAPI;
+        return res.getVariablePrivateData(id).then((result) => getEditorResponseData<null>(result));
```
Matthiee, line 619:
```suggestion
        return res.getVariablePrivateData(id).then((result) => getEditorResponseData<PrivateData>(result));
```

Matthiee (changes requested):

packages/sdk/src/controllers/VariableController.ts
brapoprod: should the setter also not return it?

packages/sdk/src/types/VariableTypes.ts
brapoprod: Could be handy

packages/sdk/src/controllers/VariableController.ts
Dvergar: Looking at other setters, it seems we mostly (only?) return when an id is generated from an entity we created. Here the setter is a full replace of the private data so I don't think it has much value.

packages/sdk/src/controllers/VariableController.ts
```diff
@@ -596,6 +597,29 @@ export class VariableController {
         return this.applyPrefixSuffixDeltaUpdate(id, update);
     };
 
+    /**
+     * Sets the private data for any variable
+     * @param id the id of the variable to update
+     * @param privateData the private data
+     * @returns
+     */
+    setPrivateData = async (id: string, privateData: Record<string, string>) => {
```
Matthiee, line 606:
```suggestion
    setPrivateData = async (id: string, privateData: PrivateData) => {
```

Matthiee (approved):
LGTM

Matthiee (approved):

PR #511, opened 2024-08-29
[Fix] Teach genie about line breaks

Dvergar:
This PR teaches genie that line breaks are not allowed on short text variable values.

## PR Guidelines

## Related tickets

https://chilipublishintranet.atlassian.net/browse/EDT-1633

### DIFF
--- a/packages/actions/src/ActionHelpers.ts
+++ b/packages/actions/src/ActionHelpers.ts
@@ -152,7 +152,7 @@ function getImageVariableValue(variableName: string | Variable): string {
 /**
  * Sets the value of a variable by its name or variable object.
  *
- * If the value is set on a ShortTextVariable, it cannot contain any type of
+ * If the value is set on a ShortTextVariable, it must not contain any type of
  * line breaks.
  *
  * @param {string | Variable} variableName - The name of the variable to update.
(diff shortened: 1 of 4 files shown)


NicolaVerbeeck (approved):

olegsolodukhin (dismissed):

packages/actions/src/ActionHelpers.ts
```diff
@@ -152,6 +152,9 @@ function getImageVariableValue(variableName: string | Variable): string {
 /**
  * Sets the value of a variable by its name or variable object.
  *
+ * If the value is set on a ShortTextVariable, it cannot contain any type of
```
evan-mcgeek, line 155: This might be more explicit
```suggestion
 * If the value is set on a ShortTextVariable, it must not contain any type of
```

Matthiee (changes requested):
This keeps generating new lines 😢

packages/actions/src/ActionHelpers.ts
```diff
@@ -162,6 +165,9 @@ function setVariableValue(variableName: string | Variable, value: VariableValue)
 /**
  * Sets the value of a text variable by its name or variable object.
  *
+ * If the value is set on a ShortTextVariable, it cannot contain any type of
```
Matthiee, line 168: The contextual data being send is ```"name": "qr", "type": "shortText"```. 

It might be better to mention "shortText" instead?

```suggestion
 * If the value is set on a variable of type shortText, it cannot contain any type of
```

_But even with this change it will create a single line string with "\n" inside of it._

Matthiee (changes requested):

packages/actions/src/ActionHelpers.ts
Matthiee: wdyt @brapoprod ?

brapoprod (approved):

Matthiee (approved):

PR #516, opened 2024-09-06
[Feature][PROPOSAL] Anchor to page

brapoprod:
This PR creates a proposal for anchor to page properties, set and reset

## PR Guidelines

### DIFF
--- a/packages/sdk/src/types/FrameTypes.ts
+++ b/packages/sdk/src/types/FrameTypes.ts
@@ -21,6 +21,7 @@ export type FrameLayoutType = {
     maxCopyfitting: PropertyState<number>;
     enableCopyfitting: PropertyState<boolean>;
     autoGrow: AutoGrowSettings;
+    anchoring: AnchorSettings;
 } | null;
 
 //Frame.image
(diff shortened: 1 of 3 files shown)


packages/sdk/src/types/FrameTypes.ts
```diff
@@ -274,3 +276,84 @@ export enum UpdateZIndexMethod {
...
+}
+
+export class FrameAnchorTarget {
+    id: Id;
+    edge: AnchorTargetEdgeType;
+    type = AnchorTargetType.frame;
+
+    constructor(id: Id, edge: AnchorTargetEdgeType) {
+        (this.id = id), (this.edge = edge);
+    }
```
brapoprod, line 310: why is this relevant?

Dvergar: Could be a misunderstanding of typescript on my side. I made it a class just so that I could create a page anchor target without having to define the type, which is already known in the type definition (& I assume is more of a constraint here). I basically wanted to do `new PageAnchorTarget()`;

Dvergar: @alexandraFlavia9 @abdelhalimkhouas WDYT?

abdelhalimkhouas: I would go with the simpler approach and just do
````

 export interface FrameAnchorTarget {
    id: Id;
    edge: AnchorTargetEdgeType;
    type = AnchorTargetType.frame;
}
````
I see u r using `new FrameAnchorTarget` in the tests, if there is no other use I suggest to use an interface instead of a class.

Dvergar: After a little chat w/ Halim we're gonna keep it as-is because this model, while it will be used from engine to SDK, it will **also** be used as an object from SDK to engine via the `setHorizontalAnchor` & `setVerticalAnchor` methods.

PR #530, opened 2024-10-29
[NO-TICKET] nullable filters and sorting in PageConfig

evan-mcgeek:
This PR makes filters and soring nillable

## PR Guidelines

### DIFF
--- a/packages/sdk/src/types/DataConnectorTypes.ts
+++ b/packages/sdk/src/types/DataConnectorTypes.ts
@@ -5,8 +5,8 @@ export type DataConnectorCapabilities = {
 };
 
 export type PageConfig = {
-    filters: DataFilter[];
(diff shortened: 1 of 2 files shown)


packages/sdk/src/types/DataConnectorTypes.ts
```diff
@@ -5,8 +5,8 @@ export type DataConnectorCapabilities = {
 };
 
 export type PageConfig = {
-    filters: DataFilter[];
-    sorting: DataSorting[];
+    filters?: DataFilter[] | null;
```
psamusev, line 8: should we update the type in studio-connectors package as well?

evan-mcgeek: we should indeed

psamusev (approved):

PR #532, opened 2024-11-12
[Feature] Add analyze method and media feature types to MediaConnector

pietervp:
This PR partially implements tech design https://chilipublishintranet.atlassian.net/wiki/spaces/dev/pages/4713250843/MediaConnector+-+Analyze+Asset+Features+SmartCrop+Captions+etc 

## PR Guidelines

## Related tickets

-  [GRAFX-3944](https://chilipublishintranet.atlassian.net/browse/GRAFX-3944)

[GRAFX-3944]: https://chilipublishintranet.atlassian.net/browse/GRAFX-3944?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/connectors/src/MediaConnector.ts
+++ b/packages/connectors/src/MediaConnector.ts
@@ -9,8 +9,32 @@ export interface MediaConnector {
         intent: DownloadIntent,
         context: Dictionary,
     ): Promise<ArrayBufferPointer>;
+    analyze(id: string, features: MediaFeatureRequest[], context: Dictionary): Promise<MediaFeatures>
+
     getConfigurationOptions(): ConnectorConfigValue[] | null;
     getCapabilities(): MediaConnectorCapabilities;
+}
+
+export const enum MediaFeatureRequest {
+    smartCrop = 'smartCrop',
+}
+
+export type MediaFeatures = {
+    subject?: MediaSubject,
+    poi?: MediaPointOfInterest,
+    category?: string,
+}
+
+export type MediaPointOfInterest = {
+    x: number;
+    y: number;
+}
+
+export type MediaSubject = {
+    x: number;
+    y: number;
+    width: number;
+    height: number;
 }
 
 export type DownloadIntent = 'web' | 'print' | 'animation';
@@ -43,4 +67,5 @@ export type MediaConnectorCapabilities = {
     detail: boolean;
     filtering: boolean;
     metadata?: boolean;
+    analyzeSmartCrop?: boolean;
 };
--- a/packages/connectors/src/MediaConnector.ts
+++ b/packages/connectors/src/MediaConnector.ts
@@ -20,9 +20,13 @@ export const enum MediaFeatureRequest {
 }
 
 export type MediaFeatures = {
-    subject?: MediaSubject,
-    poi?: MediaPointOfInterest,
-    category?: string,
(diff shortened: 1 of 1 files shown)


packages/connectors/src/MediaConnector.ts
```diff
@@ -9,10 +9,34 @@ export interface MediaConnector {
...
 }
 
+export const enum MediaFeatureRequest {
+    smartCrop = 'smartCrop',
+}
+
+export type MediaFeatures = {
+    subject?: MediaSubject,
+    poi?: MediaPointOfInterest,
+    category?: string,
```
pietervp, line 25: Create a subproperty 'smartCrop' with all those properties

packages/connectors/src/MediaConnector.ts
```diff
@@ -9,10 +9,34 @@ export interface MediaConnector {
         intent: DownloadIntent,
         context: Dictionary,
     ): Promise<ArrayBufferPointer>;
+    analyze(id: string, features: MediaFeatureRequest[], context: Dictionary): Promise<MediaFeatures>
```
psamusev, line 12: do we assume that we run run this function not only for `smartCrop` ?

packages/connectors/src/MediaConnector.ts
```diff
@@ -43,4 +67,5 @@ export type MediaConnectorCapabilities = {
     detail: boolean;
     filtering: boolean;
     metadata?: boolean;
+    analyzeSmartCrop?: boolean;
```
psamusev, line 70: if we assume that `analyze` is going to be executed not only for smartCrop, I would consider it as object instead, like
```
analyze?: {
 smartCrop: boolean
}
```
```

Comment (NicolaVerbeeck) [reply packages/connectors/src/MediaConnector.ts:70]: I feel like nesting it is not required

Comment (brapoprod) [reply packages/connectors/src/MediaConnector.ts:74]: but now it's not in sync with the capabilities...

Comment (NicolaVerbeeck) [reply packages/connectors/src/MediaConnector.ts:74]: Sure it is 🤔

**Review by brapoprod — APPROVED** [review]

**Review by psamusev — APPROVED** [review]

File: packages/connectors/src/MediaConnector.ts [hunk 4]
```diff
@@ -43,4 +71,5 @@ export type MediaConnectorCapabilities = {
...
```
Matthiee, line 74: I don't understand why we need to add `smart crop` in the capabilities, considering `analyze` method also doesn't specify it. 

```suggestion
    analyze?: boolean;
```

NicolaVerbeeck: Because we will have more things to come in analyze method later

Matthiee: Yeah after checking the engine implementation I now understand _why_. 

However this is not really in line with other capability names, was this done on purpose? 

Considering `filtering` is also not called `queryFiltering` and we make a split between `query` and `filtering` capability.

```suggestion
    analyze?: boolean; // the analyze method exists
    smartCrop?: boolean; // smart crop feature is available in analyze
```

PR #534, opened 2024-11-20
[Feature] Add data source methods

Dvergar:
No strong opinion between `onDataSourceIdChanged` and `onDataSourceChanged`

## PR Guidelines

## Related tickets

https://chilipublishintranet.atlassian.net/browse/EDT-1711
https://chilipublishintranet.atlassian.net/browse/WRS-2077

### DIFF
--- a/packages/sdk/src/controllers/ConnectorController.ts
+++ b/packages/sdk/src/controllers/ConnectorController.ts
@@ -197,7 +197,7 @@ export class ConnectorController {
      * @param id the id of your registered connector you want to make sure it is loaded
      * @returns
      */
-    private waitToBeReady = async (id: Id, timeoutMilliseconds = 2000): Promise<EditorResponse<null>> => {
+    waitToBeReady = async (id: Id, timeoutMilliseconds = 2000): Promise<EditorResponse<null>> => {
         // minimum timeout is 500ms
         let timeout = Math.max(timeoutMilliseconds, 500);
 
--- a/packages/sdk/src/controllers/ConnectorController.ts
+++ b/packages/sdk/src/controllers/ConnectorController.ts
@@ -190,6 +190,47 @@ export class ConnectorController {
     };
 
     /**
+     * Sets the data source by id.
+     * @param connectorId the id of your data connector
+     * @returns
+     */
+    setDataSource = async (connectorId: string) => {
+        const res = await this.#editorAPI;
+        return res.setDataSource(connectorId).then((result) => getEditorResponseData<null>(result));
+    };
+
+    /**
+     * Gets the data source.
+     * @returns the id of your data connector
+     */
+    getDataSource = async () => {
+        // add backwards compatibility code
+        const res = await this.#editorAPI;
+        return res
+            .getDataSource()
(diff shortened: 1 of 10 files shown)


packages/sdk/src/controllers/ConnectorController.ts
```diff
@@ -189,6 +189,47 @@ export class ConnectorController {
...
+            .then((resp) => {
+                const update: EditorResponse<ConnectorInstance> = { ...resp, parsedData: null };
+                if (resp.parsedData) {
+                    update.parsedData = this.#connectorCompatibilityTools.makeSingleConnectorBackwardsCompatible(
+                        resp.parsedData,
+                        this.#localConfig.get(WellKnownConfigurationKeys.GraFxStudioEnvironmentApiUrl),
+                    );
+                }
+                return update;
+            });
```
Matthiee, line 221: DOes it need to be backwards compatible? Considering it is a new feature.

Dvergar: I admit I did not know as I wanted to be safe there; I now just checked the compatibility source and indeed, shouldn't be a problem to remove it.

packages/sdk/src/controllers/ConnectorController.ts
```diff
@@ -189,6 +189,47 @@ export class ConnectorController {
         return res.getConnectorOptions(id).then((result) => getEditorResponseData<ConnectorOptions>(result));
     };
 
+    /**
+     * Sets the data source by id.
+     * @param connectorId the id of your data connector
+     * @returns
+     */
+    setDataSource = async (connectorId: string) => {
```
psamusev, line 197: I'm not sure that having such methods in ConnectorController is a good idea. First of all because it's related to data Connector only, where this controller created to work with any type of connector. In addition, if I'm not mistaken, there where discussion to have it as new controller, called `dataSource`

Dvergar: Fine with a data source controller tbh.

Logic of avoiding the data controller (which I considered first) is that they are all those type specific connector controllers are related to direct interaction with the running controller (js) maybe that's me thinking too much with implementation in mind and does not matter so much for the user.

I think data source controller is probably less confusing anyway.

packages/sdk/src/controllers/ConnectorController.ts
```diff
@@ -197,7 +197,7 @@ export class ConnectorController {
      * @param id the id of your registered connector you want to make sure it is loaded
      * @returns
      */
-    waitToBeReady = async (id: Id, timeoutMilliseconds = 2000): Promise<EditorResponse<null>> => {
+    private waitToBeReady = async (id: Id, timeoutMilliseconds = 2000): Promise<EditorResponse<null>> => {
```
brapoprod, line 200: in theory this is a breaking change

psamusev: I would say practically for those who are TypeScript based (We're btw, ) It's a breaking change (it's even breaking change for us in Studio UI)

psamusev (dismissed):

Matthiee (approved):

PR #549, opened 2024-12-19
[EDT-1765] SnapshotSettings introduced

evan-mcgeek:
This PR adds a new method `getSnapshotWithSettings()` with `SnapshotSettings` object to be able to control snapshot properties, eg. increase/decrease its resolution

## PR Guidelines

## Related tickets

-   [EDT-1765](https://chilipublishintranet.atlassian.net/browse/EDT-1765)

[EDT-1765]: https://chilipublishintranet.atlassian.net/browse/EDT-1765?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


evan-mcgeek:
> Why isn't `getSnapshot` kept with a nullable settings parameter instead?

good idea indeed

Dvergar (approved):
nit, I would maybe describe what is the behavior of a null `largestAxisSize` and if there is an enforced limit from our side.

packages/sdk/src/types/PageTypes.ts
```diff
@@ -13,3 +13,7 @@ export type PageSize = {
     width: number;
     height: number;
 };
+
+export interface SnapshotSettings {
+    largestAxisSize?: number | null;
```
brapoprod, line 18:
```suggestion
    maxDimensionSize?: number | null;
```

evan-mcgeek: it has been merged on the engine side, so this change would require to make changes on the engine side as well

packages/sdk/src/controllers/PageController.ts
```diff
@@ -102,12 +102,18 @@ export class PageController {
...
+     * If it's not provided, default settings will be applied, e.g., page resolution.
+     * Limit for `largestAxisSize` is 1000 px
      * @returns UInt8Array snapshot of the given page
      */
-    getSnapshot = async (pageId: Id) => {
+    getSnapshot = async (pageId?: Id, settings?: SnapshotSettings | null) => {
         const res = await this.#blobAPI;
-        return res.getPageSnapshot(pageId).then((result) => (result as Uint8Array) ?? (result as EditorResponse<null>));
+        return res
+            .getPageSnapshotWithSettings(pageId, settings)
```
Matthiee, line 115: Don't you need to send the settings as JSON?

packages/sdk/src/types/PageTypes.ts
brapoprod: so this means you'll do it or this means it's not possible? I really think it's a non descriptive vague name

evan-mcgeek: I would leave it as it is for now, considering the current code

Matthiee (approved):

packages/sdk/src/types/PageTypes.ts
Matthiee: So what are we going to do now? We can always add a comment explaining what it means.

cc @brapoprod @evan-mcgeek

brapoprod: I guess that's the least we should do idd

Dvergar (approved):

Matthiee (dismissed):

Matthiee (approved):

packages/sdk/src/controllers/PageController.ts
```diff
@@ -102,12 +102,18 @@ export class PageController {
...
+     * If it's not provided, default settings will be applied, e.g., page resolution.
+     * Limit for `largestAxisSize` is 1000 px
      * @returns UInt8Array snapshot of the given page
      */
-    getSnapshot = async (pageId: Id) => {
+    getSnapshot = async (pageId?: Id, settings?: SnapshotSettings | null) => {
         const res = await this.#blobAPI;
-        return res.getPageSnapshot(pageId).then((result) => (result as Uint8Array) ?? (result as EditorResponse<null>));
+        return res
+            .getPageSnapshotWithSettings(pageId, JSON.stringify(settings))
```
Matthiee, line 115: Only perform `JSON.stringify` if `settings` are defined

Matthiee: getSnapshotWithSettings('0', 'null') will throw

Dvergar (approved):

PR #592, opened 2025-03-20
[Feature][EDT-1888] Update text models to include text stroke

NicolaVerbeeck:
This PR also updates the models to include some missing properties

## PR Guidelines

## Related tickets

-   [EDT-1888](https://chilipublishintranet.atlassian.net/browse/EDT-1888)

## Screenshots

[EDT-1888]: https://chilipublishintranet.atlassian.net/browse/EDT-1888?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/sdk/src/types/ParagraphStyleTypes.ts
+++ b/packages/sdk/src/types/ParagraphStyleTypes.ts
@@ -88,7 +88,7 @@ export type ParagraphStyleUpdate = {
         value: boolean;
     };
     strokeWidth: {
-        value: string;
+        value: number;
     };
     underline: {
         value: boolean;
--- a/packages/sdk/src/types/ParagraphStyleTypes.ts
+++ b/packages/sdk/src/types/ParagraphStyleTypes.ts
@@ -88,7 +88,7 @@ export type ParagraphStyleUpdate = {
         value: boolean;
     };
     strokeWidth: {
-        value: number;
+        value: string;
     };
     underline: {
         value: boolean;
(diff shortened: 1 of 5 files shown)


Matthiee (approved):

Matthiee (dismissed):

brapoprod (dismissed):

Dvergar (dismissed):

packages/sdk/src/types/ParagraphStyleTypes.ts
```diff
@@ -75,6 +78,18 @@ export type ParagraphStyleUpdate = {
...
+        value: boolean;
+    };
+    strokeColor: {
+        value: ColorUsageUpdate;
+    };
+    strokeColorApplied: {
+        value: boolean;
+    };
+    strokeWidth: {
+        value: string;
```
Dvergar, line 91: What was the constraint on `strokeWidth` and points i'm confused, we default the measurement type to points on the engine side when not defined in the string itself but this is just a regular unit otherwise 🤔

NicolaVerbeeck: For updating we allow strings so the end user can write formulas or have the 'pt' attached etc. We always send points back

Dvergar: I assumed string was by default considered unitValue by SDK but if that's not the case then engine side dtos need to be updated as they are sending double right now.

Dvergar (approved):

psamusev (approved):

PR #595, opened 2025-03-28
[Fix] Add UAT deploy (version-management only)

brapoprod:
This PR adds the UAT deploy step, for version management only

### DIFF
--- a/.github/workflows/update-engine-version.yml
+++ b/.github/workflows/update-engine-version.yml
@@ -11,6 +11,14 @@ on:
         description: 'New engine version (e.g. 2.1.1)'
         required: true
         type: string
+      environment:
+        description: 'Target environment (UAT or PROD)'
+        required: true
+        type: choice
+        default: 'UAT'
+        options:
+          - 'UAT'
+          - 'PROD'
 
 jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/update-engine-version.yml
```diff
@@ -65,4 +73,28 @@ jobs:
         uses: azure/CLI@v1
         with:
           inlineScript: |
-            az storage blob upload --file temp/supported-versions.json --container-name shared --name version-management/supported-versions.json --connection-string "${{ secrets.AZURE_CDN_STUDIO_PRD_CONNECTION_STRING }}" --overwrite true 
\ No newline at end of file
+            az storage blob upload --file temp/supported-versions.json --container-name shared --name version-management/supported-versions.json --connection-string "${{ secrets.AZURE_CDN_STUDIO_PRD_CONNECTION_STRING }}" --overwrite true
```
NicolaVerbeeck, line 76: Isn't this uploading twice now, once always to prod and once to the target env?

brapoprod: you're right

NicolaVerbeeck (approved):

PR #596, opened 2025-03-31
[Fix] Allow event skipping w/ custom undo

Dvergar:
This PR allows setting a custom undo operation without triggering an event.
This method is in experimental as we want to find a more long-term approach to that scenario.

## PR Guidelines

## Related tickets

https://chilipublishintranet.atlassian.net/browse/EDT-1902
https://chilipublishintranet.atlassian.net/browse/MAIN-1156

### DIFF
--- a/packages/sdk/src/controllers/ExperimentController.ts
+++ b/packages/sdk/src/controllers/ExperimentController.ts
@@ -101,4 +101,20 @@ export class ExperimentController {
         const res = await this.#editorAPI;
         return res.selectTextById(frameId, startIndex, length).then((result) => getEditorResponseData<null>(result));
     };
+
+    /**
+     * This method adds custom data that will be saved and restored when undoing and redoing.
+     * Duplicate values are overwritten. The data is exposed via the onCustomUndoDataChanged event.
+     *
+     * @param key The key of the custom data
+     * @param value The value of the custom data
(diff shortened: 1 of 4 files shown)


Matthiee (approved):

packages/sdk/src/controllers/ExperimentController.ts
```diff
@@ -101,4 +101,20 @@ export class ExperimentController {
...
+    /**
+     * This method adds custom data that will be saved and restored when undoing and redoing.
+     * Duplicate values are overwritten. The data is exposed via the onCustomUndoDataChanged event.
+     *
+     * @param key The key of the custom data
+     * @param value The value of the custom data
+     * @param skipEvent Whether to skip the event that is triggered when the custom data is changed, defaults to false
+     * @returns
+     */
+    addCustomUndoData = async (key: string, value: string, skipEvent?: boolean) => {
```
psamusev, line 114: We've already have similar method under`undoManager` controller called `addCustomData`. What's the difference here? Can we simply extend existing method rather than introducing a new one?

Dvergar: We're not super fond of the idea of adding a super specific argument such as that one, which is currently targeted at one client only. We'd like to think about another approach but for now we decided to unblock the situation by adding it in the experimental controller.

psamusev (approved):

PR #605, opened 2025-04-30
[Feature] Media connector upload

brapoprod:
This PR

## PR Guidelines

- [ ] I have read the [contribution and PR guidelines](/chili-publish/studio-sdk/blob/main/CONTRIBUTING.md)

## Related tickets

-   [WRS-NUMBER](https://support.chili-publish.com/projects/WRS/issues/WRS-NUMBER)


abdelhalimkhouas (dismissed):

packages/connectors/src/MediaConnector.ts
```diff
@@ -9,6 +9,7 @@ export interface MediaConnector {
         intent: DownloadIntent,
         context: Dictionary,
     ): Promise<ArrayBufferPointer>;
+    upload(filePointers: FilePointer[], context: Dictionary): Promise<Media[]>;
```
psamusev, line 12: It will enforce existing connectors to implement the method, although it should be considered as optional. Can we define it as separate interface instead?

packages/connectors/src/Connector.Shared.ts
```diff
@@ -12,6 +12,13 @@ export interface ConnectorRuntimeContext {
     sdkVersion: string;
 }
 
+export interface FilePointer { 
+    id: string;
+    url: string;
+    name: string;
+    type: 'filePointer';
```
psamusev, line 19: what's the purpose of this?

packages/connectors/src/Connector.Shared.ts
```diff
@@ -21,11 +28,23 @@ export type QueryOptions = {
     sortBy: string | null;
 };
 
+export interface StudioFormData {
+    append(name: string, value: string | FilePointer): void;
+    set(name: string, value: string | FilePointer): void;
+    forEach(callback: (value: string | FilePointer, name: string) => void): void;
+    getFields(): Array<[string, string | FilePointer]>;
```
psamusev, line 35: does it correctly state for array of arrays?

packages/connectors/src/Connector.Shared.ts
```diff
@@ -69,11 +88,13 @@ export interface ChiliResponse extends ChiliBody {
 }
 
 export type ConnectorConfigValueType = 'text' | 'boolean';
+export type ConnectorConfigContextType = 'query' | 'upload';
 
 export interface ConnectorConfigValue {
     readonly name: string;
     readonly displayName: string;
     readonly type: ConnectorConfigValueType;
+    readonly context: ConnectorConfigContextType[];
```
psamusev, line 97: what's the purpose of adding a new field? Wouldn't it be simpler to extend existing `ConnectorConfigValueType` with `upload` type?

psamusev (changes requested):

packages/connectors/src/Connector.Shared.ts
abdelhalimkhouas: this serves as a pointer to the file to upload, first files will be in a staging state then we will perform the upload, this will hold the reference to the file.

packages/connectors/src/Connector.Shared.ts
abdelhalimkhouas: but the `ConnectorConfigContextType` is the type of the action or the operational context, not the data type represented by `ConnectorConfigValueType`, or at least that's what i understood

packages/connectors/src/Connector.Shared.ts
psamusev: @abdelhalimkhouas  Thx, for the explanation, but I rather meant the `type` field of the FilePointer interface :)

packages/connectors/src/Connector.Shared.ts
psamusev, line 97: Should be optional to avoid breaking changes of existing connectors

packages/connectors/src/Connector.Shared.ts
NicolaVerbeeck: It is there to ensure that if we get a random json object to upload, we can more safely identify that it is a file pointer instead of random json (which we need to know for BE)

packages/connectors/src/MediaConnector.ts
NicolaVerbeeck: Could be. @brapoprod I am guessing that this will work for our cases

packages/connectors/src/Connector.Shared.ts
NicolaVerbeeck: I don't understand the question

packages/connectors/src/MediaConnector.ts
brapoprod: how do you see this @psamusev ?

packages/connectors/src/Connector.Shared.ts
brapoprod: flutter will handle this, if it's undefined they will treat it as browse. I'm wondering if this can be a breaking change since it's inside the connector package?

packages/sdk/src/controllers/UtilsController.ts
```diff
@@ -26,4 +35,54 @@ export class UtilsController {
         const host = type == EnvironmentType.SANDBOX ? 'chili-publish-sandbox' : 'chili-publish';
         return `https://${environment}.${host}.online/grafx/api/v${version}/environment/${environment}`;
     };
+
+    /**
+     * This method can help you stage a file to the CHILI GraFx Environment API for upload.
+     * @param files The Files or Blobs to stage.
+     * @returns Promise<FilePointer[]> referencing the staged data.
+     */
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
brapoprod, line 44: remove

packages/connectors/src/MediaConnector.ts
psamusev: Separate interface with method definition, i.e.
```typescript
export interface MediaConnectorUpload {
   upload(filePointers: FilePointer[], context: Dictionary): Promise<Media[]>;
}
```

and its usage

```typescript
 class Connector implements Media.MediaConnector {...} // current use case, no changes

class ConnectorWithUpload implements Media.MediaConnector, Media.MediaConnectorUpload {...} // explicitly specify we wanna implement upload, requires connector to implement the upload method
```

packages/connectors/src/Connector.Shared.ts
psamusev: @NicolaVerbeeck can we then create a private type, i.e. PrivateFilePointer and use it there (for example how we did with Date type for Data Connector). I have a concern to exposing it in public interface, since it doesn't have any sense when using in the connector upload method

packages/connectors/src/Connector.Shared.ts
psamusev: shouldn't it be `getFields(): Array<string | FilePointer>` return type?

packages/connectors/src/Connector.Shared.ts
psamusev: It's a breaking change from API definition - it enforce existing connectors to define context as empty array in their ConnectorConfigValues

psamusev, line 97:
```suggestion
    readonly context?: ConnectorConfigContextType[];
```

packages/connectors/src/Connector.Shared.ts
psamusev: @NicolaVerbeeck can we then create a private type, i.e. PrivateFilePointer and use it there (for example how we did with Date type for Data Connector). I have a concern to exposing it in public interface, since it doesn't have any sense when using in the connector upload method

packages/connectors/src/Connector.Shared.ts
psamusev: should it has `Array<string | FilePointer>` as return type instead?

packages/connectors/src/Connector.Shared.ts
psamusev: I mean it's a breaking change from this interface definition, should be define following in such case

```suggestion
readonly context?: ConnectorConfigContextType[];
```

psamusev (dismissed):

psamusev (dismissed):

Matthiee (approved):

packages/sdk/src/controllers/UtilsController.ts
```diff
@@ -26,4 +35,53 @@ export class UtilsController {
...
+
+        const response = await fetch(stageUrl, {
+            method: 'POST',
+            body: formData,
+            headers: {
+                'Authorization': `Bearer ${accessToken}`,
+            },
+        });
+
+        if (!response.ok) {
```
psamusev, line 78: Here we should have a logic to throw specific errors based on backend response, according to Product ticket

psamusev (approved):

packages/sdk/src/controllers/UtilsController.ts
psamusev: Fixed

psamusev (approved):

NicolaVerbeeck (approved):

PR #608, opened 2025-05-08
[PRODUCT-714] Add min upload dimensions on image variable

brapoprod:
This PR adds the minimum upload dimensions capability on upload to the image variable settings

## PR Guidelines

## Related tickets

-   [PRODUCT-714](https://support.chili-publish.com/projects/WRS/issues/PRODUCT-714)

### DIFF
--- a/packages/sdk/src/types/VariableTypes.ts
+++ b/packages/sdk/src/types/VariableTypes.ts
@@ -55,6 +55,8 @@ export interface ImageVariable extends Variable {
     value?: ConnectorImageVariableSource;
     allowQuery: boolean;
     allowUpload: boolean;
+    uploadMinHeight?: number;
+    uploadMinWidth?: number;
 }
 
 export interface ListVariableItem {
--- a/packages/sdk/src/controllers/VariableController.ts
+++ b/packages/sdk/src/controllers/VariableController.ts
@@ -680,6 +680,28 @@ export class VariableController {
         return res.setImageVariableAllowUpload(id, allowUpload).then((result) => getEditorResponseData<null>(result));
     };
 
+    /**
+     * This method sets the minimum height for an image variable that will be uploaded
+     * @param id the id of the variable
+     * @param minHeight the minimum height
+     * @returns
+     */
+    setImageUploadMinHeight = async (id: string, minHeight: number) => {
+        const res = await this.#editorAPI;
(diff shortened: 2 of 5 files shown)


NicolaVerbeeck (approved):

Dvergar (approved):
The ticket on our side is at the very bottom of the backlog, should that be sorted/re-prioritized? @NicolaVerbeeck

NicolaVerbeeck:
> The ticket on our side is at the very bottom of the backlog, should that be sorted/re-prioritized? @NicolaVerbeeck

Fixed

psamusev (dismissed):

Dvergar (changes requested):
EDT-1962 suggests to also have one call to set both components.

> Width and height can be set separately but we can use a single SDK call for this

~~If there is such change, I'd prefer a redirection to a `res.setImageVariableUploadMinSize(id, null, minHeight)` for the specific calls.~~

Edit: wait, the ticket suggests null as a remove, let me think

psamusev (approved):

Dvergar:
> > setImageVariableUploadMinSize
> 
> There is no reset method necessary afaik. Resetting should be done by assigning null

If resetting is mandatory to be done by assigning `null` I'd suggest some changes: As I see from your last commit, `null` is used as a _"not gonna set"_ meaning (on the api call) so we cannot reset this way.

We could have what I proposed on my previous comment or maybe this would be more elegant, pseudo-code:

```typescript
    setMinImageUploadWidth = async (id: string, minWidth: number | null) => {
        const res = await this.#editorAPI;
        if(minWidth == null) {
            return  res.resetImageVariableUploadMinSize(id, true, false...
        }
        return res.setImageVariableUploadMinSize(id, minWidth, null).then((result) => getEditorResponseData<null>(result));
    };
```

packages/sdk/src/types/VariableTypes.ts
```diff
@@ -55,6 +55,8 @@ export interface ImageVariable extends Variable {
     value?: ConnectorImageVariableSource;
     allowQuery: boolean;
     allowUpload: boolean;
+    uploadMinHeight?: number;
+    uploadMinWidth?: number;
```
Dvergar, line 59:
```suggestion
    uploadMinHeight?: number | null;
    uploadMinWidth?: number | null;
```

psamusev (dismissed):

Dvergar (approved):

psamusev (approved):

PR #616, opened 2025-05-20
[Fix] Change editor domain on uat

brapoprod:
This PR cahnges release flow on UAT and adds an editor domain


NicolaVerbeeck (approved):
I guess

PR #766, opened 2026-04-16
[NO-TICKET] Added container to ImageFrame

asrbljaninchili:
This PR

## PR Guidelines

## Related tickets

- [NO-TICKET]

### DIFF
--- a/packages/sdk/src/types/FrameTypes.ts
+++ b/packages/sdk/src/types/FrameTypes.ts
@@ -80,7 +80,7 @@ export type ImageFrame = {
     src?: ImageFrameSource;
     blendMode: BlendMode;
     dropShadowSettings?: DropShadowSettings;
-    container?: {
+    container: {
         source: ShapeFrameSource;
         properties: ShapeContainerProperties;
     }
--- a/packages/sdk/src/types/FrameTypes.ts
+++ b/packages/sdk/src/types/FrameTypes.ts
@@ -3,7 +3,7 @@ import { BarcodeCharacterSet, BarcodeErrorCorrectionLevel, BarcodeType, QuietZon
 import { ColorUsage } from './ColorStyleTypes';
 import { HasOverrideState, Id, PropertyState } from './CommonTypes';
 import { GradientUsage } from './GradientStyleTypes';
-import { CornerRadiusAll, CornerRadiusNone, CornerRadiusOnly, ShapeType } from './ShapeTypes';
+import { CornerRadiusAll, CornerRadiusNone, CornerRadiusOnly, ShapeFrameSource, ShapeType, ShapeContainerProperties } from './ShapeTypes';
 
 export type FrameLayoutType = {
     id: Id;
@@ -80,6 +80,10 @@ export type ImageFrame = {
     src?: ImageFrameSource;
     blendMode: BlendMode;
     dropShadowSettings?: DropShadowSettings;
+    container: {
+        source: ShapeFrameSource;
+        properties: ShapeContainerProperties;
+    }
 };
 
 export type ShapeFrame = {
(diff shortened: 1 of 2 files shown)


NicolaVerbeeck (dismissed):

abdelhalimkhouas (dismissed):

packages/sdk/src/types/FrameTypes.ts
```diff
@@ -80,6 +80,10 @@ export type ImageFrame = {
     src?: ImageFrameSource;
     blendMode: BlendMode;
     dropShadowSettings?: DropShadowSettings;
+    container?: {
+        source: ShapeFrameSource;
+        properties: ShapeContainerProperties;
+    }
```
Dvergar:
```suggestion
    container: {
        source: ShapeFrameSource;
        properties: ShapeContainerProperties;
    }
```
I know this is matching engine but i'm wondering why it's nullable on the engine side, given that this object always exists 👀

packages/sdk/src/types/ShapeTypes.ts
```diff
@@ -15,6 +15,13 @@ export interface ShapeProperties {
     allCornersSame?: boolean;
 }
 
+export interface ShapeContainerProperties {
+    enableStroke?: boolean;
+    strokeColor?: ColorUsage;
+    strokeWeight?: number;
+    allCornersSame?: boolean;
```
Dvergar: Those are always defined on engine.

packages/sdk/src/types/ShapeTypes.ts
```diff
@@ -60,8 +67,8 @@ export type ShapeFrameEllipse = {
 
 export type ShapeFramePolygon = {
     type: ShapeType.polygon;
-    sides: number;
+    sides?: number;
```
Dvergar, line 70: This is defined on engine side.

packages/sdk/src/types/ShapeTypes.ts
```diff
@@ -60,8 +67,8 @@ export type ShapeFrameEllipse = {
...
     cornerRadius?: CornerRadius;
 };
 
-export type ShapeFrameSource = ShapeFrameRect | ShapeFrameEllipse | ShapeFramePolygon;
\ No newline at end of file
+export type ShapeFrameSource = ShapeFrameRect | ShapeFrameEllipse | ShapeFramePolygon;
```
Dvergar:
```suggestion
export type ShapeFrameSource = ShapeFrameRectSource | ShapeFrameEllipseSource | ShapeFramePolygonSource;
```

Dvergar (changes requested):

packages/sdk/src/types/ShapeTypes.ts
asrbljaninchili: Yes, but in this case, WRS would only need to set the type and not sides property so I guess that is why @abdelhalimkhouas made the change

Use case: setting the shape type of the image frame

packages/sdk/src/types/ShapeTypes.ts
asrbljaninchili: It is and at the moment it is only used for WRS to read these properties. For updating properties, we use ShapeProperties so there is no need for these properties to be optional. I will change it

packages/sdk/src/types/FrameTypes.ts
asrbljaninchili: You are right, changing on engine as well ✍️

Dvergar (approved):

packages/sdk/src/types/ShapeTypes.ts
abdelhalimkhouas: Correct @asrbljaninchili

abdelhalimkhouas (approved):

PR #812, opened 2026-07-31
[AIML-40] Fix memory leak of run mode

brapoprod:
## Related tickets

-   [AIML-40](https://chilipublishintranet.atlassian.net/browse/AIML-40)

## Problem

A heap snapshot taken after opening and closing run mode twice contained **three complete, fully live engines** — 674MB of reachable heap, growing ~250MB per run-mode visit. There is no engine cache involved: each visit builds a brand new engine and none of the old ones are ever released.

This PR adds the missing teardown primitive. It is the SDK half of the fix; the consumer halves are in studio-ui and studio-workspace.

## Root cause in the SDK

`Connect` created the engine iframe but never stored it anywhere retrievable — the element was only reachable from closures. In particular, the module-level `messageHandler` closes over the iframe:

```ts
let messageHandler = null;
// ...
messageHandler = (event) => { if (event.source !== iframe.contentWindow) return; /* ... */ };
window.addEventListener('message', messageHandler);
```

As long as that listener stays on `window`, the iframe element is reachable, and a referenced iframe keeps its **entire realm** alive. There was no `SDK.destroy()`, so a consumer had no way to release any of it.

## Evidence from the snapshot

Per engine instance: a 32MB CanvasKit `HEAP8`, two 16MB QuickJS runtimes, and a dart2js heap whose library holder exclusively retained **128.2MB**. Counted across the snapshot:

| | live | leaked |
|---|---|---|
| `flutter-view` / `canvas` | 1 | 2 |
| `<iframe>` (with layout object) | 2 | 2 |
| objects with a `HEAP8` property | 3 | 6 |
| `FiberRootNode` (vs. 1 `HostRoot`) | 1 | 2 |

## Changes

- Track the engine iframe and the `DOMContentLoaded` handler at module level, and expose `teardownFrame()` to release them.
- Add **`SDK.destroy()`**: destroys the penpal connection, removes the iframe, clears the module-level connection. Idempotent, and tolerant of a connection penpal already tore down via its own iframe-removal monitor.
- Run the same teardown at the start of `Connect`, so a second `loadEditor()` can no longer orphan the previous engine.
- Keep `this.connection` in sync in `setConnection`. It was assigned in the constructor, which runs *before* `loadEditor`, so it pointed at the previous editor's connection (or nothing on a first load) — `destroy()` needs the live one.

## Notes for reviewers

- `connection` is typed non-optional, so clearing the module-level reference needs a cast rather than a public type change. Controllers are deliberately left in place: calls after `destroy()` reject with penpal's "destroyed connection" error, which is the sane behaviour.
- `teardownFrame()` at the top of `Connect` replaces the old `removeEventListener` call, so the listener handling is strictly a superset of what was there.

## Verification

- `yarn test` — 642 passed, 51 suites
- `yarn ci-lint` — 0 errors (3 pre-existing warnings)
- `tsc --noEmit` — clean
- Two new tests cover the teardown: the iframe and its `message` listener are released, and connecting a second time releases the first frame.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[AIML-40]: https://chilipublishintranet.atlassian.net/browse/AIML-40?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/sdk/src/sdk.ts
+++ b/packages/sdk/src/sdk.ts
@@ -1,7 +1,7 @@
 import { Connection } from 'penpal';
 import engineInfo from '../editor-engine.json';
 import packageInfo from '../package.json';
-import Connect, { teardownFrame } from './interactions/Connector';
+import Connect from './interactions/Connector';
 import { defaultStudioOptions, WellKnownConfigurationKeys } from './types/ConfigurationTypes';
 
 import type { ConfigType, EditorAPI, RuntimeConfigType } from './types/CommonTypes';
@@ -55,8 +55,6 @@ declare const __ENGINE_DOMAIN__: string;
 const ENGINE_DOMAIN = typeof __ENGINE_DOMAIN__ !== 'undefined' ? __ENGINE_DOMAIN__ : 'studio-cdn.chiligrafx.com';
 const FIXED_EDITOR_LINK = `https://${ENGINE_DOMAIN}/editor/${engineInfo.current}/web`;
 
-let connection: Connection;
-
 // Marks a freshly created controller as "needs instrumentation" without requiring
 // its property name to be tracked anywhere. `applyMethodInstrumentation` later scans
 // the SDK instance for properties carrying this marker and wraps them automatically,
@@ -77,7 +75,8 @@ const isPendingInstrumentation = (value: unknown): value is object => {
 
 export class SDK {
     config: RuntimeConfigType;
-    connection: Connection;
+    // Set by `loadEditor` via `setConnection`; stays `undefined` until then.
+    connection!: Connection;
 
     /**
      * @ignore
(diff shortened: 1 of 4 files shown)


Matthiee (dismissed):

alexandraFlavia9 (dismissed):

packages/sdk/src/sdk.ts
```diff
@@ -341,6 +341,39 @@ export class SDK {
 
     setConnection = (newConnection: Connection) => {
         connection = newConnection;
```
psamusev: do we need both, module level and constructor field reference for connection?

alexandraFlavia9 (approved):

NicolaVerbeeck (approved):

